Issue #112, opened 2023-11-10
make yarn lint pass


paddymul:
Yarn lint currently fails locally (and presumably in CI)

paddymul:
fixed by #113

Issue #110, opened 2023-11-10
Make existing pytest pass


paddymul:
The pytest tests fail in CI with the following errors

```
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED tests/auto_clean_test.py::test_get_typing_metadata - AssertionError: assert {'bool': 0.0,...or': 1.0, ...} == {'bool': 0.0,...or': 0.0, ...}
  Omitting 7 identical items, use -vv to show
  Differing items:
  {'datetime_error': 1.0} != {'datetime_error': 0.0}
  Full diff:
    {
     'bool': 0.0,
     'bool_error': 1.0,
     'datetime': 1.0,
  -  'datetime_error': 0.0,
  ?                    ^
  +  'datetime_error': 1.0,
  ?                    ^
     'float': 0.0,
  -  'float_error': 1.0,
  ?                  --
  +  'float_error': 1,
     'int': 0.0,
     'int_error': 1.0,
    }
FAILED tests/basic_widget_test.py::test_basic_instantiation - NotImplementedError: <class 'numpy.uint8'>
FAILED tests/basic_widget_test.py::test_word_only_instantiation - NotImplementedError: <class 'numpy.uint8'>
FAILED tests/basic_widget_test.py::test_basic_display - NotImplementedError: <class 'numpy.uint8'>
FAILED tests/basic_widget_test.py::test_interpreter - NotImplementedError: <class 'numpy.uint8'>
FAILED tests/basic_widget_test.py::test_interpreter_errors - NotImplementedError: <class 'numpy.uint8'>
FAILED tests/basic_widget_test.py::test_analysis_pipeline - NotImplementedError: <class 'numpy.uint8'>
FAILED tests/basic_widget_test.py::test_auto_clean_true - NotImplementedError: <class 'numpy.uint8'>
FAILED tests/basic_widget_test.py::test_post_processing - NotImplementedError: <class 'numpy.uint8'>
```
This works on my mac

paddymul:
fixed by #113

Issue #108, opened 2023-11-10
State of CI brain dump


paddymul:
The CI Build for buckaroo is currently broken, and never completely worked.  Building buckaroo (a pre-req for most CI steps is very complex).  Buckaroo has regular python code (tested via pytest without much fanfare), typescript code (without tests, but it should pass typechecking), and an opaque build process that depends on obscure python build packages that also somehow install npm.   I wrote a blog post about my initial struggles getting this project to build https://medium.com/@paddy_mullen/writing-a-jupyter-widget-my-experience-78c40f8033c7 .  There are multiple bug reports in that post.

1. When building locally I use https://github.com/paddymul/buckaroo/blob/main/full_build.sh
2. I can succesfully build a wheel that is installable on my mac.
3. This uses the python `build` module which interacts with parts of https://github.com/paddymul/buckaroo/blob/main/pyproject.toml to load extra jupyter extension modules.
4. These extension modules set up some type of virtual machine or clean environment.  I really don't know, they are very opaque with the actual commands being run.  I will paste the output of `python -m build .` as a comment
5. Relevant parts of pyproject.toml
```

[tool.hatch.build]
artifacts = [
    "buckaroo/nbextension/index.*",
    "buckaroo/labextension",
]
```
```
[tool.hatch.build.targets.wheel.shared-data]
"buckaroo/nbextension" = "share/jupyter/nbextensions/buckaroo"
"buckaroo/labextension" = "share/jupyter/labextensions/buckaroo"
"buckaroo.json" = "etc/jupyter/nbconfig/notebook.d/buckaroo.json"
```
```
[tool.hatch.build.hooks.jupyter-builder]
```
```
build-function = "hatch_jupyter_builder.npm_builder"
```

6. Current CI status.  I stopped looking at CI status months ago because it wasn't helpful.  Now it is apparently failing on python unit tests.  This should be a fairly easy fix that I can tackle.  Previously it failed on setting up the proper npm environment from python for building.
Most recent failure
```
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED tests/auto_clean_test.py::test_get_typing_metadata - AssertionError: assert {'bool': 0.0,...or': 1.0, ...} == {'bool': 0.0,...or': 0.0, ...}
  Omitting 7 identical items, use -vv to show
  Differing items:
  {'datetime_error': 1.0} != {'datetime_error': 0.0}
  Full diff:
    {
[...]

PR #113, opened 2023-11-10
Chore/fix pytest

paddymul:
Release 0.4.1 - Fix Continuous Integration.
* Fixes:  State of CI brain dump #108 
* Fixes:  Make existing pytest pass #110 
* Fixes:  Make yarn lint pass #112 
* Fixes:  Hatchling CI error #114 
* Fixes:  CI-build only typescript typing errors #115

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,4 @@ dist
 coverage
 **/*.d.ts
 tests
+js/components/RechartTooltip.tsx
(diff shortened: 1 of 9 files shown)


.eslintignore
```diff
@@ -2,4 +2,5 @@ node_modules
 dist
 coverage
 **/*.d.ts
-tests
\ No newline at end of file
+tests
+js/components/RechartTooltip.tsx
```
paddymul, line 6: ignoring all of RechartTooltip.tsx because it is wholesale copied from Recharts codebase to override specific functionality, that codebase had different tslint standards, and the code works.

PR #154, opened 2023-11-16
Feat/python submodules

paddymul:


buckaroo/auto_clean/cleaning_commands.py
```diff
@@ -1,5 +1,5 @@
-from .lispy import s
-from .configure_utils import configure_buckaroo
+from ..jlisp.lispy import s
```
ghost, line 1: unable to import ..jlisp.lispy. it throws me the following error
ImportError: attempted relative import with no known parent package

PR #175, opened 2023-11-23
Feat/pluggable analytics framework split series ops

### LINKED ISSUE #171: Separate series ops from imperative ops in Pluggable Analysis Framework
### Checks

- [X] I have checked that this enhancement has not already been requested

### How would you categorize this request.  You can select multiple if not sure

Summary stats, Performance, Developer Experience/CI (feature to make it easier to devlop on Buckaroo)

### Enhancement Description

For PAF most of the runtime is take up by series operations (nlargest, mean, median)

Most of the business logic is in regular python code. Look at assembling histograms as an example.

### Pseudo Code Implementation

Refactor PAF so it works as follows, for polars in particular

have a `series_selector` function that returns statements like
```python
class NewAnalysis:
    provides_series = ["mean", "bottom", "top"]
    @static_method
    def series_selector():
        return [
                    F.all().mean().name.prefix("mean:")
                    F.col(pl.Int64, pl.UInt32).bottom_k(5).name.prefix("bottom:"),
                    F.col(pl.Int64, pl.UInt32).top_k(5).name.prefix("top:")]
```

Then pluggable analysis framework works to make the rest of the logic work with regular python dag stuff

This way we can implement pandas logic and polars series methods, and have the rest of the python code work the same.

### Prior Art

N/A

paddymul:
The current PAF (Pluggable Analysis Framework) has a single function for `ColAnalysis` objects - `summary(sampled_ser, summary_ser, ser)`, which accretively builds summary_ser dicts.  

This splits `ColAnalysis` into two functions `series_summary(sampled_ser, ser)`  `computed_summary(summary_dict)`.

`computed_summary` does the regular python logic, `series_summary` is the only function that gets access to the the full series.  I think this separation will make it easier to plug in polars.  For polars, `series_summary` will be implemented differently.

Most of the bugs with analysis code come from weird series interactions which then return an unexpected `NaN`.

Fixes #171  
Separate series ops from imperative ops in Pluggable Analysis Framework


paddymul:
superseeded by #182 .  Includes everything from this branch and more.

PR #271, opened 2024-04-12
trying to upgrade build environment to jupyterlab 4.0

paddymul:


paddymul:
This branch worked as a nice test run for #273 which fixed the problem

PR #310, opened 2024-10-18
try1 , pointing to lib instead of dist

paddymul:


paddymul:
Moving away from webpack, so this is irrelevant

PR #333, opened 2024-11-14
trying to point to lib/index.js

paddymul:


paddymul:
Not relevant because #339 moves away from webpack

PR #334, opened 2024-11-16
buckaroo widget displays without styles vis ipyreact

paddymul:


paddymul:
This was the basis for work in #339 .  #339 works, this doesn't

PR #345, opened 2025-01-28
Feat/msg infinite scrolling

paddymul:
This branch sets up infinite scrolling via callbacks from a custom message, this should have less latency than the previous polling of react state variables.  Code should end up cleaner too


paddymul:
included in  #353 . closed

PR #467, opened 2026-01-11
Pin uv version to 0.5.15 to speed up CI

paddymul:
## Summary
- Pin uv to version 0.5.15 in all CI jobs to enable better caching

## Background
The "Install uv" step in CI was taking longer than expected. By pinning to a specific version, GitHub Actions can cache the uv binary more effectively and avoid checking for the latest version on every run.

## Test plan
- Monitor CI run times, particularly the "Install uv" step
- Verify all jobs still pass

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     # runs-on: depot-ubuntu-24.04
     strategy:
       fail-fast: false
-       matrix:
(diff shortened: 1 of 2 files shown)


paddymul:
Testing optimized uv installer - should be much faster now

paddymul:
this was not actually faster,  closing the branch

PR #530, opened 2026-02-21
fix: debug and fix WASM marimo test selectors (approach A)

paddymul:
## Summary
- **Root cause**: `micropip.install("buckaroo", keep_going=True)` fails in Pyodide because `fastparquet` (a C extension) has no pure-Python wheel. Despite `keep_going=True`, the install raises a `ValueError`, crashing the first cell and cascading failures to all downstream cells that use BuckarooWidget.
- **WASM notebook fix**: Install buckaroo's pure-Python deps explicitly, then install buckaroo with `deps=False`, and create a minimal `fastparquet` stub module (`types.ModuleType`) so buckaroo's import chain succeeds. Applied to all three WASM notebooks.
- **Python fix**: Make `fastparquet` import optional in `buckaroo/serialization_utils.py` via `try/except ImportError`, so future releases don't need the stub workaround.
- **Test fix**: Replace `waitFor` with fixed timeout (which was too short at 60s) with a polling loop that waits up to 240s for content + 60s for widgets, with console log capture for CI debugging.

## Test plan
- [x] Ran WASM marimo Playwright test locally 3 times -- all pass in ~18-22s
- [ ] CI should pass with the updated WASM HTML (needs `bash scripts/marimo_wasm_output.sh buckaroo_simple.py run` in CI or pre-committed HTML)
- [ ] Verify existing Python tests still pass (fastparquet import is backward-compatible)

### DIFF
--- a/buckaroo/serialization_utils.py
+++ b/buckaroo/serialization_utils.py
@@ -5,8 +5,12 @@ import pandas as pd
 from typing import Dict, Any, List, Tuple
 from pandas._libs.tslibs import timezones
 from pandas.core.dtypes.dtypes import DatetimeTZDtype
-from fastparquet import json as fp_json
 import logging
+
+try:
+    from fastparquet import json as fp_json
+except ImportError:
+    fp_json = None  # fastparquet not available (e.g. Pyodide/WASM)
 
 from buckaroo.df_util import old_col_new_col, to_chars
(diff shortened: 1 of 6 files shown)


paddymul:
Closing in favor of #532 (uses real Pyodide-bundled fastparquet)

PR #539, opened 2026-02-22
fix: inherit pinned row displayer uses column formatter

paddymul:
## Summary
- Fix the `inherit` pinned row displayer to correctly apply the column's own formatter
- Instead of trying to extract `valueFormatter` from AG-Grid's `colDef` at runtime (which doesn't work for pinned rows), pass `column_config` into `getCellRendererSelector` and build a `col_name → formatter` lookup map
- The `inherit` handler uses this map to apply the column's displayer to pinned row values

## What was broken
Pinned rows with `displayer: "inherit"` showed raw unformatted values (e.g. `894.8674` instead of `895` for an integer column, `15634.842223667903` instead of `1,052` for std).

## What this fixes
- Integer column: `mean=894.8674` → `895` (uses the column's `float(0)` displayer)
- Float column: `mean=3.14159` → `3.142` (uses the column's `float(3)` displayer)

## Test plan
- [x] TypeScript compiles clean
- [x] Added storybook story (`InheritPinnedRows`) with int + float columns
- [x] Added playwright test (`inherit-pinned-rows.spec.ts`) verifying formatting
- [ ] CI passes (storybook playwright tests)


paddymul:
redundant from #545

PR #548, opened 2026-02-23
refactor: abstract CustomizableDataflow into pandas/polars subclasses

paddymul:
## Summary
- Makes `CustomizableDataflow` an abstract base class with 4 abstract methods: `_compute_processed_result`, `_build_error_dataframe`, `_get_summary_sd`, `_df_to_obj`
- Creates `PandasCustomizableDataflow` (new file) and `PolarsCustomizableDataflow` (new file) as concrete implementations
- Widgets select their backend via a `dataflow_klass` class attribute instead of inheriting from the pandas widget
- Removes pandas imports from `dataflow.py` and `dataflow_extras.py` base classes

## Motivation
This decouples the dataflow pipeline from pandas, enabling pandas to become an optional dependency. Previously `PolarsBuckarooWidget` inherited from the pandas widget and overrode methods — now each backend has its own clean dataflow subclass.

## Files changed
| File | Change |
|------|--------|
| `buckaroo/dataflow/dataflow.py` | Make `CustomizableDataflow` abstract, remove pandas imports |
| `buckaroo/dataflow/dataflow_extras.py` | Remove pandas imports, duck-type `sort_index` check |
| `buckaroo/dataflow/pandas_dataflow.py` | **New**: `PandasCustomizableDataflow` |
| `buckaroo/dataflow/polars_dataflow.py` | **New**: `PolarsCustomizableDataflow` |
| `buckaroo/buckaroo_widget.py` | Add `dataflow_klass`, use it in `InnerDataFlow` |
| `buckaroo/polars_buckaroo.py` | Set `dataflow_klass = PolarsCustomizableDataflow` |
| `buckaroo/server/data_loading.py` | Use `PandasCustomizableDataflow` |
| `tests/unit/dataflow/*.py` | Update imports to use `PandasCustomizableDataflow` |

## Test plan
- [x] All 586 unit tests pass locally (5 skipped)
- [ ] CI passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,3 +276,4 @@ Have you had a good experience with this project? Why not share some love and co
 We welcome [issue reports](../../issues); be sure to choose the proper issue template for your issue, so that we can be sure you're providing the necessary information.
 
 
+
--- a/buckaroo/buckaroo_widget.py
+++ b/buckaroo/buckaroo_widget.py
@@ -27,7 +27,7 @@ from .pluggable_analysis_framework.col_analysis import ColAnalysis
 from buckaroo.extension_utils import copy_extend
 
 from .serialization_utils import EMPTY_DF_WHOLE, check_and_fix_df, pd_to_obj, to_parquet, sd_to_parquet_b64
-from .dataflow.dataflow import CustomizableDataflow
+from .dataflow.pandas_dataflow import PandasCustomizableDataflow
 from .dataflow.dataflow_extras import (Sampling, exception_protect)
 from .dataflow.styling_core import (ComponentConfig, DFViewerConfig, DisplayArgs, OverrideColumnConfig, PinnedRowConfig, StylingAnalysis, merge_column_config, EMPTY_DFVIEWER_CONFIG)
 from .dataflow.autocleaning import PandasAutocleaning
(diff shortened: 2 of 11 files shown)


paddymul:
## Learnings & Self-Review

### Was it an improvement?

**Structurally, yes.** The core win is that `dataflow.py` and `dataflow_extras.py` no longer import pandas. This is the prerequisite for making pandas an optional dependency. Before this PR, `PolarsBuckarooWidget` inherited from `BuckarooWidget` (a pandas widget) and monkey-patched methods like `_df_to_obj` and `_build_error_dataframe`. Now each backend has a clean subclass.

The `dataflow_klass` attribute pattern is cleaner than the old `InnerDataFlow` approach that delegated `_df_to_obj` back to the widget via closure — that mixed widget concerns into the dataflow layer.

**However, there's significant code duplication.** `_compute_processed_result` and `_get_summary_sd` are identical between `PandasCustomizableDataflow` and `PolarsCustomizableDataflow`. Only `_build_error_dataframe` (`pd.DataFrame` vs `pl.DataFrame`) and `_df_to_obj` (direct vs `.to_pandas()`) actually differ. A follow-up could keep only those two as abstract and hoist the shared logic back to the base class.

### Performance implications around serialization

**No change for pandas users.** The `_df_to_obj` path is identical — `pd_to_obj(self.sampling_klass.serialize_sample(df))`.

**The polars path is unchanged too** — `PolarsCustomizableDataflow._df_to_obj` does the same `df.to_pandas()` conversion that `PolarsBuckarooWidget._df_to_obj` did before. The infinite scroll path (`_handle_payload_args`) writes polars directly to parquet and was not touched.

**One subtle behavioral change in `dataflow_extras.py`:** The base `Sampling.pre_stats_sample` used to call `sampled.sort_index()` on pandas DataFrames after sampling. That was removed from the base class (no pandas import), but `PdSampling` in `buckaroo_widget.py` already overrides `pre_stats_sample` with its own `sort_index()` call, so pandas widget path is unaffected. `serialize_sample` was changed from `isinstance(sampled, pd.DataFrame)` to `hasattr(sampled, 'sort_index')` — duck typing that works without importing pandas.

### What this PR doesn't do yet

- Doesn't make pandas optional at import time — `serialization_utils.py` and many other files still `import pandas` at module level
- `PolarsCustomizableDataflow` still imports pandas (for the `isinstance` check in `_df_to_obj`)
- `autocleaning.py` still defines `SENTINEL_DF_1`/`SENTINEL_DF_2` with pandas — used in production code
- `_compute_processed_result` and `_get_summary_sd` are copy-pasted between both subclasses — follow-up should hoist shared logic

paddymul:
@codex

PR #549, opened 2026-02-23
refactor: switch transport from Parquet to Arrow IPC

paddymul:
## Summary

Replace `fastparquet` (Python) + `hyparquet` (JS) with `pyarrow.ipc` + `@uwdata/flechette` for binary data transport between Python and JS.

- Drop `fastparquet` from core dependencies — key enabler for making pandas optional and for WASM/Pyodide compatibility
- Synchronous JS decoding with `tableFromIPC()` eliminates the async/sync callback duality from hyparquet
- Backend-agnostic serialization: works with both pandas (`pa.Table.from_pandas()`) and polars (`df.to_arrow()`)
- Backward compatibility aliases preserved (`to_parquet = to_arrow_ipc`, etc.)

---

### JS bundle size impact

| Bundle | main | This PR | Delta |
|--------|------|---------|-------|
| `widget.js` (Jupyter) | 3,514,857 B | 3,519,203 B | **+4,346 B (+0.12%)** |

The switch from hyparquet (~10 KB min+gz) to flechette (~14 KB min+gz) adds roughly **+4 KB** to the final widget bundle — negligible.

### Python dependency impact

| Dependency | Install size | Status |
|------------|-------------|--------|
| `fastparquet` | 4.1 MB | **Removed from core deps** |
| `cramjam` (fastparquet transitive) | 7.8 MB | **No longer pulled in** |
| `pyarrow` | 113 MB | Already a hard dep (unchanged) |

Net savings: **~12 MB** removed from install footprint. More importantly, `cramjam` is a compiled Rust extension with no WASM wheel, so removing it unblocks Pyodide/WASM deployment.

### Decoding architecture: why Arrow IPC is better for widget transport

**Arrow IPC streaming format** is fundamentally different from Parquet for this use case:

| | Arrow IPC | Parquet |
|---|---|---|
| **Decode model** | Buffer reinterpretation (near-zero cost) | Decompress → reverse encoding → assemble row groups |
| **JS API** | `tableFromIPC(buf)` — **synchronous** | `parquetRead({file, onComplete})` — **async only** |
| **Python encode cost** | `ipc.new_stream()` — trivial, Arrow-native | `fastparquet.write()` — compression + dictionary/RLE encoding |
| **Wire size** | Larger (uncompressed by default) | Smaller (compressed) |
| **Best for** | In-process IPC, widget transport | Remote file access, HTTP range requests |

The wire size tradeoff doesn't matter here — data travels over a local Jupyter websocket or kernel comm channel, not the internet. The decode speed and API simplicity advantages dominate.

### Concrete wins from synchronous decoding

The old hyparquet code path had a subtle bug surface area because `parquetRead` fires its `onComplete` callback asynchronously in some bundler environments (esbuild standalone) but synchronously in others (webpack/Jupyter). This forced us to maintain both `resolveDFData()` (sync, returns `[]` if callback hasn't fired yet) and `resolveDFDataAsync()` (async, wraps in Promise) plus a `preResolveDFDataDict()` pre-resolution step.

With flechette, `tableFromIPC()` is always synchronous. `resolveDFDataAsync` now trivially wraps the sync version, and `preResolveDFDataDict` no longer needs `Promise.all`. This eliminates an entire class of timing bugs where summary stats would briefly render as empty arrays before the async decode completed.

### Row extraction performance

From flechette's published benchmarks (vs the official `apache-arrow` JS library):

| Operation | Flechette speedup |
|-----------|------------------|
| Row object extraction (`table.toArray()`) | **7–11x faster** |
| Array extraction | 2–7x faster |
| Value iteration | 1.3–1.6x faster |

We use `table.toArray()` (row object extraction) on every infinite scroll response and every summary stats decode — this is the hot path. The 7–11x speedup over `apache-arrow` JS is comparing Arrow libraries; the advantage over hyparquet's Parquet decode path is even larger since Parquet decoding has additional decompression overhead on top of extraction.

---

### Python (8 files)
- `serialization_utils.py`: `to_arrow_ipc()` and `sd_to_ipc_b64()` replace fastparquet-based functions
- All callers updated: `buckaroo_widget`, `polars_buckaroo`, `lazy_infinite_polars_widget`, `geopandas_buckaroo`, `dataflow/dataflow`, `dataflow/column_executor_dataflow`, `server/data_loading`

### JS/TS (7 files)
- `resolveDFData.ts`: synchronous `tableFromIPC()` replaces async `parquetRead`
- `BuckarooWidgetInfinite.tsx`: flechette for infinite scroll buffer parsing
- `DFWhole.ts`: added `IpcB64Payload` type to union
- `index.ts`, `widget.tsx`, `standalone.tsx`: updated imports and transcript recording

### Config
- `pyproject.toml`: removed `fastparquet` from core `dependencies`
- `package.json`: replaced `hyparquet` with `@uwdata/flechette`

### Tests (4 files)
- Updated to use `to_arrow_ipc`, `_read_ipc_to_polars` / `_read_ipc_buffer` helpers
- Updated `_resolve_all_stats` to handle `ipc_b64` format

## Test plan

- [x] `uv run pytest tests/unit/serialization_utils_test.py -v` — serialization round-trip
- [x] `uv run pytest tests/unit -v` — 514 tests pass
- [ ] `cd packages/buckaroo-js-core && pnpm test` — JS unit tests
- [ ] Full build with `./scripts/full_build.sh`
- [ ] Manual: Jupyter notebook with infinite scroll widget


paddymul:
@codex

PR #587, opened 2026-02-25
Fix left-pinned index column for summary stats

paddymul:
## Summary
- Fixed bug where `left_col_configs` columns weren't actually pinned left — return statement used `lcc2` instead of `lcc3` (which has `pinned: 'left'` applied)
- Removed leftover `console.log("lcc3", lcc3)` debug statement
- The index/stat label column now stays visible when scrolling horizontally through wide DataFrames

Supersedes #541 (rebased cleanly on current main).

## Test plan
- [x] All 80 jest tests pass
- [ ] Verify in Storybook stories with `left_col_configs` that the index column stays pinned left on horizontal scroll


paddymul:
@codex

PR #589, opened 2026-02-25
Fix BuckarooCompare for arbitrary join keys

### LINKED ISSUE #588: issues with the compare tool
### Checks

- [x] I have checked that this enhancement has not already been requested

### How would you categorize this request.  You can select multiple if not sure

Other (please describe in Issue Description)

### Enhancement Description

# GitHub Bug Report Draft: `BuckarooCompare` is not generic across join keys/schemas

## Title
Bug: `BuckarooCompare` assumes join column `a` and mis-handles `join_columns`, breaking generic use

## Type
Bug

## Affected area
- Buckaroo compare demo utility
- File: `docs/example-notebooks/marimo-wasm/buckaroo_compare.py`

## Summary
`BuckarooCompare` currently contains hard-coded and inconsistent join-key logic that causes incorrect behavior when used outside the demo schema.

This should be filed as a **bug against buckaroo**, because the function signature implies generic behavior (`join_columns`, `how`) but the implementation relies on fixed assumptions (notably `a`) and partial list/scalar key handling.

## Current behavior (actual)

1. Membership logic is hard-coded to `a`:
   - `df_1_membership = m_df['a'].isin(df1[join_columns])`
   - `df_2_membership = (m_df['a'].isin(df2[join_columns]).astype('Int8') *2)`

2. `join_columns` is inconsistently treated as both list and scalar:
   - `BuckarooCompare(..., join_columns, ...)` calls `col_join_dfs(..., join_columns[0], ...)`.
   - `col_join_dfs` then passes `join_columns` into `pd.merge(..., on=join_columns, ...)`.

3. Column-level diff counts are computed pre-merge via positional comparison:
   - `(df1[col] != df2[col]).sum()`
   - This can be wrong when row order differs or rows exist only on one side.

## Expected behavior
- The compare utility should work with **any join-key name** and **multi-column keys**.
- The type and semantics of `join_columns` should be consistent end-to-end.
- Diff summaries should be computed from **keyed-aligned merged rows**, not index position.

### Repro 1: non-`a` join key
1. Use two dataframes keyed by `id` (no `a` column).
2. Call `BuckarooCompare(df1, df2, join_columns=['id'], how='outer')`.

Expected: comparison works.
Actual: membership logic referencing `m_df['a']` fails.

### Repro 2: multi-key join
1. Use two dataframes with join keys `['account_id', 'as_of_date']`.
2. Call `BuckarooCompare(df1, df2, join_columns=['account_id', 'as_of_date'], how='outer')`.

Expected: a valid two-key compare.
Actual: `join_columns[0]` drops the second key, causing incorrect behavior.

## Impact
- Prevents generic reuse in real datasets.
- Produces incorrect summaries for common data quality scenarios (reordered rows, added/removed rows, multi-key joins).
- Creates mismatch between public API shape and actual behavior.

## Proposed fix
1. Normalize `join_columns` as `list[str]` and validate once at API boundary.
2. Remove hard-coded `a`; compute membership via merge indicator (`indicator=True`) or key-presence logic derived from `join_columns`.
3. Align rows by keys first; compute column diff statistics after merge.
4. Keep demo notebook, but extract reusable compare logic into a core module for testing.
5. Add tests covering:
   - non-`a` key
   - multi-key joins
   - reordered rows
   - one-sided extra rows
   - null-heavy comparisons

## Acceptance criteria
- `BuckarooCompare` succeeds for arbitrary key names and multi-key joins.
- No schema-specific hard-coded assumptions remain in comparison logic.
- Diff stats are derived from aligned merged rows.
- Tests cover regression cases above and pass in CI.

## References
- Implementation under test: `docs/example-notebooks/marimo-wasm/buckaroo_compare.py`

### Pseudo Code Implementation

_No response_

### Prior Art

n/a

paddymul:
## Summary
- Extracts `col_join_dfs` from the marimo demo notebook into `buckaroo/compare.py` as a reusable, tested module
- Fixes hardcoded `m_df['a']` membership logic — now uses `pd.merge(indicator=True)` to work with any join key name
- Fixes inconsistent `join_columns` handling — now always `list[str]` end-to-end (no more `join_columns[0]` unwrap)
- Computes diff stats from key-aligned merged rows instead of positional comparison
- Fixes `column_config_overrides[join_columns]` (was passing list as dict key) — now applies config per join column

## Test plan
- [x] 9 new tests in `tests/unit/compare_test.py` covering:
  - Non-`a` join key
  - Multi-key joins (`account_id`, `as_of_date`)
  - Outer join membership (df1-only, df2-only, both)
  - Reordered rows (key-aligned diff correctness)
  - One-sided extra columns
  - String `join_columns` normalization
  - Sentinel column name rejection
  - Inner join
  - Null-heavy comparisons
- [x] All 9 tests pass locally

Closes #588

### DIFF
--- a/buckaroo/compare.py
+++ b/buckaroo/compare.py
@@ -1,0 +1,117 @@
+import pandas as pd
+
+
+def col_join_dfs(df1, df2, join_columns, how):
+    """Join two DataFrames and compute column-level diff statistics.
+
+    Parameters
+    ----------
+    df1, df2 : pd.DataFrame
+        The two DataFrames to compare.
+    join_columns : str or list[str]
+        Column name(s) to join on.
+    how : str
+        Join type passed to ``pd.merge`` (e.g. 'inner', 'outer', 'left', 'right').
+
+    Returns
+    -------
+    m_df : pd.DataFrame
+        Merged DataFrame with membership and equality columns.
+    column_config_overrides : dict
+        Buckaroo column config for styling.
+    eqs : dict
+        Per-column diff summary.
+    """
+    # Normalize join_columns to list
+    if isinstance(join_columns, str):
+        join_columns = [join_columns]
+
+    df2_suffix = "|df2"
+    for col in df1.columns:
+        if df2_suffix in col:
+            raise ValueError(
+                f"|df2 is a sentinel column name used by this tool, "
+                f"and it can't be used in a dataframe passed in, {col} violates that constraint"
+            )
+    for col in df2.columns:
+        if df2_suffix in col:
+            raise ValueError(
+                f"|df2 is a sentinel column name used by this tool, "
+                f"and it can't be used in a dataframe passed in, {col} violates that constraint"
+            )
+
+    df1_name, df2_name = "df_1", "df_2"
+
+    # Merge first so diff stats are computed on key-aligned rows
+    m_df = pd.merge(
+        df1, df2, on=join_columns, how=how, suffixes=["", df2_suffix], indicator=True
+    )
+
+    # Compute membership from merge indicator
+    # 1 = df1 only, 2 = df2 only, 3 = both
+    membership_map = {"left_only": 1, "right_only": 2, "both": 3}
+    m_df["membership"] = m_df["_merge"].map(membership_map).astype("Int8")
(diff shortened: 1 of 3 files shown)


paddymul:
@codex

PR #686, opened 2026-04-30
Add ibis/xorq analysis backend (v2 @stat pipeline)

paddymul:
## Summary

Replaces #518 (closed because the GitHub UI showed a phantom 499-file diff after a clean rebase — actual diff is 10 files).

Adds an ibis-backed analysis backend to buckaroo, ported to the v2 `@stat` / `StatPipeline` framework following an adversarial review of the original v1-style implementation.

### What's in
- **`IbisStatPipeline`** (`buckaroo/pluggable_analysis_framework/ibis_stat_pipeline.py`) — two-phase executor: a single batched `table.aggregate(...)` query for all `IbisColumn`-param @stat funcs, then per-column post-batch (computed stats + `IbisTable`-param histograms) through the standard typed-DAG executor.
- **`ibis_stats_v2.py`** rewritten as `@stat`-decorated functions: typing, length, null_count, min, max, distinct_count, mean, std, median, non_null_count, nan_per, distinct_per, histogram. Naming now matches `pd_stats_v2` (was v1-style classes despite the `_v2` filename).
- **`IbisDfStatsV2`** (`df_stats_v2.py`) — DfStats-shaped wrapper so DataFlow / `CustomizableDataflow` consumers can run against an `ibis.Table` without changes.
- **Markers**: `IbisColumn` and `IbisTable` added to `RAW_MARKER_TYPES` + a small `ibis_table` injection branch in `_execute_stat_func`.
- **`xorq` optional dependency group** in pyproject.toml; lock pins xorq 0.3.21 + ibis-framework 12.0.0.

### Adversarial review fixes baked in
- **Histogram is live** — was dead code (defined but never registered); numeric + categorical paths both have tests asserting populated histograms.
- **Structured error capture** — three silent `except Exception: continue` blocks replaced with `Ok`/`Err` accumulator semantics; failures surface as `StatError`. Test asserts a bad `@stat` shows up in `errors`.
- **DAG ordering** — implicit via `@stat` parameter typing; no longer a doc comment.
- **Naming v1 vs v2** — the file is `_v2` and the contents now match the `@stat` v2 system.
- **Empty-table length=0** — `SUM` over empty returns NULL in SQL; coalesced.
- **Real widget path** — IbisDfStatsV2 makes the pipeline reachable from DataFlow.

## Test plan
- [x] `ruff check` + `ruff format` clean on changed files
[...]

PR #691, opened 2026-05-01
feat: XorqStatPipeline — v2 stat pipeline for xorq table inputs

paddymul:
## Summary
Lets buckaroo's stat pipeline run against a **xorq expression** (xorq's vendored ibis) instead of materialising the table to pandas first. Stats compile to a single ``table.aggregate(...)`` query in the batch phase, plus per-column histogram queries in phase two — computation is pushed to the backend (DuckDB, Postgres, Snowflake, …) via xorq, not pulled into the local Python process.

This is a fresh PR off main replacing #686 — same feature, but trimmed to the actual xorq-substantive changes (no format noise from rebase churn, no half-finished tangents).

## What's added
- **`XorqStatPipeline`** — two-phase executor. Batch aggregate folds every XorqColumn-param stat into one query. Per-column post-batch runs computed and XorqTable-param stats (histogram). DAG validation up front; smoke-test against `xo.memtable(PERVERSE_DF)` at construction.
- **`XORQ_STATS_V2`** — typing, length/null/min/max/distinct/mean/std/median (batched), non_null/nan_per/distinct_per (computed), histogram (per-column).
- **`XorqDfStatsV2`** wrapper — DfStats-shaped surface so DataFlow / CustomizableDataflow can run xorq tables unchanged.
- **Marker types** `XorqColumn` / `XorqTable` / `XorqExecute` in `stat_func.py`; `_execute_stat_func` learned `xorq_table=` / `xorq_execute=` injection slots.
- **`@stat(provides='key')`** — override the auto-derived provides name, so `base_min(col: XorqColumn) -> float` can be named `base_min` while providing the `min` key consumed by other stats.

## Deps
- New `[xorq]` extra: `xorq>=0.1.0 ; python_version < '3.14'`. xorq 0.3.21 caps `pyarrow<22` and requires `python<3.14`, both of which conflict with the `pyarrow>=22` floor for 3.14 (#689). The gate keeps everyone happy until xorq supports 3.14.
- **No raw `ibis-framework` dependency anywhere.** `buckaroo/` source imports `xorq.api` / `xorq.vendor.ibis` only — the two ibis namespaces have distinct, incompatible Table/dtype/op classes, and mixing them would silently couple us to a second package whose identities don't match xorq's vendored copies. `test_no_raw_ibis_imports` enforces the rule as a grep test.

## Removed
- `pluggable_analysis_framework/ibis_analysis.py` — orphaned v1 attempt, not imported anywhere; replaced by `XorqStatPipeline`.

## Test plan
- [x] `pytest tests/unit/test_no_raw_ibis_imports.py tests/unit/test_xorq_stats_v2.py tests/unit/test_xorq_df_stats_v2.py` — 39 green
- [x] Full `pytest tests/unit/` — 905 passed (only the pre-existing `mp_timeout_decorator_test.py` flake and one MCP install test deselected)
- [x] `ruff check` — clean
- [x] `uv lock` — resolves cleanly across 3.11–3.14 matrix
- [ ] CI green

## Replaces
Closes #686 (older branch carried a lot of unrelated reformatting churn — easier to follow as a fresh PR; #688's paddy-format roll-out also conflicted with the older branch's ruff-format work).


buckaroo/customizations/xorq_stats_v2.py
```diff
@@ -0,0 +1,275 @@
...
+# ============================================================
+# Typing — derive type flags from the schema dtype string
+# ============================================================
+
+TypingResult = TypedDict("TypingResult",
+    {"is_numeric": bool, "is_integer": bool, "is_float": bool, "is_bool": bool, "is_datetime": bool, "is_string": bool})
+
+
+@stat()
+def typing_stats(dtype: str) -> TypingResult:
```
paddymul, line 72: Should dtype be an enum or literal?

buckaroo/customizations/xorq_stats_v2.py
```diff
@@ -0,0 +1,275 @@
...
+        return 0.0
+    return distinct_count / length
+
+
+# ============================================================
+# Histogram — per-column GROUP BY query against the ibis Table
+# ============================================================
+
+
+def _numeric_histogram(execute, table, col, min_val, max_val, total_rows):
```
paddymul: expr, not tabl.

These should all have types, and hopefully _numeric histogram is tested in isolation.

I need to check how the extra query/execute calls work.  That would mean that this isn't folded into one big ibis select call

buckaroo/customizations/xorq_stats_v2.py
```diff
@@ -0,0 +1,275 @@
...
+    if len(df) == 0:
+        return []
+    total = float(df["__count"].sum())
+    if total == 0:
+        return []
+
+    bucket_width = (max_val - min_val) / 10
+    out = []
+    for _, row in df.iterrows():
+        idx = int(row["__bucket"])
```
paddymul, line 230: this is df.iterrows, but it's only on the buckets I think, so it should be fine.

buckaroo/customizations/xorq_stats_v2.py
```diff
@@ -0,0 +1,275 @@
...
+
+    ``default=[]`` provides a graceful fallback if the histogram query
+    fails (e.g. backend rejects the GROUP BY). NB: until #687 is fixed,
+    the error is silently replaced by ``[]``.
+    """
+    if length == 0:
+        return []
+    if is_numeric and not is_bool:
+        return _numeric_histogram(execute, table, orig_col_name, min, max, length)
+    return _categorical_histogram(execute, table, orig_col_name)
```
paddymul: in other examples, if the number of numerical values is less than 5 I think, we call categorical_histogram

buckaroo/pluggable_analysis_framework/df_stats_v2.py
```diff
@@ -12,6 +12,7 @@
     stats.sdf  # -> SDType
     stats.errs  # -> ErrDict (v1 compatible)
 """
+
```
paddymul: remove,  unnecessary whitespace diff

buckaroo/pluggable_analysis_framework/df_stats_v2.py
```diff
@@ -23,6 +24,7 @@
 from .stat_pipeline import StatPipeline
 from .utils import FAST_SUMMARY_WHEN_GREATER
 from .safe_summary_df import output_full_reproduce
+from .xorq_stat_pipeline import XorqStatPipeline
```
paddymul: Does this import xorq?  we should have a CI job that tests an install with just polars - makes sure that bucakroo works properly,  just pandas, makes sure buckaroo works properly...

We don't want every install of buckaroo to require xorq to avoid import errors

paddymul: Yeah this probably shouldn't import xorq specific stuff here.  the xorq implementation should do the same thing the polars impl does

buckaroo/pluggable_analysis_framework/stat_func.py
```diff
@@ -38,43 +40,86 @@ def __bool__(self):
...
+
+
+RAW_MARKER_TYPES = (RawSeries, SampledSeries, RawDataFrame, XorqColumn, XorqTable, XorqExecute)
 
 
 # ---------------------------------------------------------------------------
 # StatKey — a named, typed slot in the DAG
 # ---------------------------------------------------------------------------
 
+
```
paddymul: lots of whitespace only diffs in this file, remove them

buckaroo/pluggable_analysis_framework/stat_func.py
```diff
@@ -150,14 +197,21 @@ def _get_requires_from_params(sig: inspect.Signature, hints: dict) -> tuple:
 # @stat decorator
 # ---------------------------------------------------------------------------
 
-def stat(column_filter=None, quiet=False, default=MISSING):
+
+def stat(column_filter=None, quiet=False, default=MISSING, provides=None):
```
paddymul: Why is provides necessary here again?

buckaroo/pluggable_analysis_framework/stat_pipeline.py
```diff
@@ -12,8 +13,8 @@
 from buckaroo.df_util import old_col_new_col
 
 from .col_analysis import ColAnalysis, ErrDict, SDType
-from .stat_func import (StatFunc, RawSeries, SampledSeries, RawDataFrame, RAW_MARKER_TYPES, MISSING, collect_stat_funcs)
-from .stat_result import Ok, Err, UpstreamError, StatError, StatResult, resolve_accumulator
+from .stat_func import (StatFunc, RawSeries, SampledSeries, RawDataFrame, XorqTable, XorqExecute, RAW_MARKER_TYPES, MISSING, collect_stat_funcs)
```
paddymul: We should always refer to XorqExpressions, not XorqTables

buckaroo/pluggable_analysis_framework/stat_pipeline.py
```diff
@@ -273,14 +286,16 @@ def process_df(self, df: pd.DataFrame, debug: bool = False) -> Tuple[SDType, Lis
...
-                initial_stats={'orig_col_name': orig_col_name, 'rewritten_col_name': rewritten_col_name})
+                initial_stats={"orig_col_name": orig_col_name, "rewritten_col_name": rewritten_col_name})
 
             summary[rewritten_col_name] = col_result
             all_errors.extend(col_errors)
 
         return summary, all_errors
 
-    def process_df_v1_compat(self, df: pd.DataFrame, debug: bool = False) -> Tuple[SDType, ErrDict]:
+    def process_df_v1_compat(
```
paddymul: Do we still have any v1 code anywhere?

buckaroo/pluggable_analysis_framework/stat_pipeline.py
```diff
@@ -399,9 +427,9 @@ def _find_v1_class(stat_func: Optional[StatFunc], original_inputs: list) -> Any:
 
     # The v1 adapter names funcs as "ClassName__series" or "ClassName__computed"
     name = stat_func.name
-    for suffix in ('__series', '__computed'):
+    for suffix in ("__series", "__computed"):
         if name.endswith(suffix):
-            class_name = name[:-len(suffix)]
+            class_name = name[: -len(suffix)]
```
paddymul: across the entire PR, I don't want to remove formatting / whitespace only diff lines.  if it fails lint because of that, I can fix separately.  I want a cleaner diff

buckaroo/customizations/xorq_stats_v2.py
paddymul: Kept as `str`. ibis dtype reprs are open-ended — parametrised forms like `decimal(10, 2)`, `timestamp('UTC')`, `array<int64>`, `struct<a: int64, b: string>` can't be enumerated up front. Prefix matching is the only practical option. Added a note to the `typing_stats` docstring spelling that out. (00d0ba53)

buckaroo/customizations/xorq_stats_v2.py
paddymul: Renamed `XorqTable` → `XorqExpr` and `table` → `expr` in both histogram helpers; added type hints (`Callable[[Any], pd.DataFrame]`, `expr: Any`, `col: str`).

The histogram queries can't fold into the batch — they `GROUP BY` on data — so they do go out as their own per-column queries (one per non-bool column). But `min`/`max` are read out of the batch accumulator, not recomputed; `test_histogram_does_not_recompute_min_max` pins this at ≤5 backend calls for a 4-column table (1 batch + 4 histogram).

Numeric histogram is exercised in isolation by `test_numeric_histogram_present` and `test_numeric_histogram_with_nulls` (regression for the null-bucket bug). (00d0ba53)

buckaroo/customizations/xorq_stats_v2.py
paddymul: Right — `df` here is the GROUP BY result, bounded at ≤10 rows for numeric (10 buckets) and ≤10 for categorical (`.limit(10)`). No row-of-data iteration.

buckaroo/customizations/xorq_stats_v2.py
paddymul: Threshold is 5 in pd_stats_v2 (`pd_stats_v2.py:217`: `len(value_counts) > 5`). Mirrored that in `histogram()` — `is_numeric and not is_bool and distinct_count > 5` → numeric, else categorical.

Bumped `_make_table()` test fixtures from 5 → 7 rows so the numeric histogram path is actually exercised by `test_numeric_histogram_present` and the backend-call counter test. Updated `test_numeric_histogram_with_nulls` to 9 distinct values for the same reason. (00d0ba53)

buckaroo/pluggable_analysis_framework/df_stats_v2.py
paddymul: Reverted. `df_stats_v2.py` is now byte-identical to main — `XorqDfStatsV2` and the xorq import moved to `xorq_stat_pipeline.py` (which already gates xorq behind `try: import xorq.api`). (00d0ba53)

buckaroo/pluggable_analysis_framework/df_stats_v2.py
paddymul: Moved `XorqDfStatsV2` to `buckaroo/pluggable_analysis_framework/xorq_stat_pipeline.py` (which already gates the `xorq.api` import behind `try/except ImportError`). `df_stats_v2.py` no longer references xorq at all — `from buckaroo.pluggable_analysis_framework.df_stats_v2 import DfStatsV2, PlDfStatsV2` works on a pandas/polars-only install.

Worth filing a follow-up to add a `pip install buckaroo` (no extras) CI job that imports each `DfStats` class — that would catch this kind of regression at the install level rather than relying on me noticing the import. (00d0ba53)

paddymul: Done — `XorqDfStatsV2` lives in `xorq_stat_pipeline.py` alongside `XorqStatPipeline`, mirroring the polars-impl-stays-out-of-pandas-module pattern. (00d0ba53)

buckaroo/pluggable_analysis_framework/stat_func.py
paddymul: Reverted. `stat_func.py` diff vs main shrank ~93 → ~63 lines, all substantive: the `XorqColumn` / `XorqExpr` / `XorqExecute` marker types, the `RAW_MARKER_TYPES` tuple update, and the `provides=` kwarg on `@stat`. (00d0ba53)

buckaroo/pluggable_analysis_framework/stat_func.py
paddymul: It lets a stat func keep a descriptive name (`base_min`) while writing into a stable accumulator key (`min`) shared with the polars/pandas implementations. Without it:

- every backend would have to define its own `min` / `max` / `distinct_count` function, and
- downstream stats would have to know which one ran (or each backend would need its own `histogram` that reads a backend-specific key).

It also avoids shadowing the `min` / `max` builtins inside the module. Expanded the `@stat` docstring with this rationale + an example. (00d0ba53)

buckaroo/pluggable_analysis_framework/stat_pipeline.py
paddymul: Renamed throughout the PR: `XorqTable` → `XorqExpr` (marker type in `stat_func.py`), `xorq_table=` kwarg in `_execute_stat_func` → `xorq_expr=`, `table` arg in histogram helpers → `expr`, plus the docstrings. (00d0ba53)

buckaroo/pluggable_analysis_framework/stat_pipeline.py
paddymul: Yes — `process_df_v1_compat` and `_find_v1_class` are still load-bearing:

- `DfStatsV2.__init__` (`df_stats_v2.py:53`), `PlDfStatsV2.__init__` (`:102`), and `XorqDfStatsV2.__init__` all call `process_df_v1_compat` / `process_table_v1_compat` to get the v1 `ErrDict` shape that `output_full_reproduce` expects.
- `_normalize_inputs` adapts `ColAnalysis` subclasses via `v1_adapter.col_analysis_to_stat_funcs` so consumers can mix v1 classes into the input list.
- `_find_v1_class` maps stat-func errors back to the original `ColAnalysis` class for the reproduce-code output.

Real consumers: `buckaroo_widget.py`, `dataflow/dataflow.py`, `dataflow/autocleaning.py`, `server/data_loading.py`, `polars_buckaroo.py`. Removing v1 means rewriting all of them.

Added a docstring note on `process_df_v1_compat` listing the callers so it's discoverable. Probably worth a separate "kill v1 surface" issue once we're ready. (00d0ba53)

buckaroo/pluggable_analysis_framework/stat_pipeline.py
paddymul: Reverted across the PR. Final `git diff main --stat` for the three previously-noisy files:

- `df_stats_v2.py`: 0 changes (whitespace + xorq import + new class — all moved out)
- `stat_func.py`: 63 lines (was 93), all substantive (XorqColumn/XorqExpr/XorqExecute + `provides=` kwarg)
- `stat_pipeline.py`: 13 lines (was 78), all substantive (XorqExpr/XorqExecute imports + branch in `_execute_stat_func` + one docstring note)

No more quote-style or blank-line-only changes anywhere in the PR. (00d0ba53)

PR #703, opened 2026-05-05
feat(xorq): XorqBuckarooWidget — buckaroo against an ibis/xorq expression

paddymul:
## Summary

- `XorqBuckarooWidget(expr)` — buckaroo over a xorq/ibis expression. Stats push down through `XorqDfStatsV2` (one batched aggregate + per-column histogram queries); only a display-sized sample lands in pandas via `expr.limit(N).execute()`.
- `add_processing(fn)` takes an expression-to-expression function so postprocessing also pushes down — filtered stats reflect the filtered expression without re-materialising.
- `BuckarooWidgetBase` gains a `dataflow_klass` class attribute so the inner dataflow can be swapped without duplicating `__init__`. Pandas widget unchanged.

## What's new in `buckaroo/xorq_buckaroo.py`

- `XorqSampling` — `serialize_sample` does `expr.limit().execute()`; pre-stats sampling is a no-op (the point is push-down).
- `XorqAutocleaning` — pass-through. The lisp interpreter targets pandas DataFrames; for ibis exprs we skip cleaning entirely.
- `XorqDataflow` — overrides `populate_df_meta` to use `expr.count().execute()` instead of `len()`, and rewrites the summary-dict keys from original column names (what `XorqStatPipeline` produces) to the rewritten `a, b, c` space the frontend works in.
- `XorqBuckarooWidget` — wires the above together with `XORQ_STATS_V2` + the default styling klasses, and exposes `add_processing(expr_func)`.

## Test plan

- [x] 9 new unit tests in `tests/unit/test_xorq_buckaroo_widget.py` cover instantiation, `df_meta` from `count()`, column rewrite, push-down filter via `add_processing`, switching back to no-op processing, and the error-frame fallback when a postprocessor raises
- [x] full unit suite (931 tests) still passes
- [x] manually verified: `XorqBuckarooWidget(memtable).add_processing(lambda e: e.filter(e.qty > 2))` → 4-row filtered display, stats min/max reflect the filtered expr (9.9 / 50.0 vs 7.4 / 50.0 unfiltered)


buckaroo/xorq_buckaroo.py
```diff
@@ -0,0 +1,193 @@
...
+"""
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+import pandas as pd
+
+from .buckaroo_widget import BuckarooWidget
+from .customizations.pl_autocleaning_conf import NoCleaningConfPl
```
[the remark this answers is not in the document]
paddymul: Already addressed in fa3cb1ca: `NoCleaningConfPl` was replaced with `NoCleaningConfXorq` (defined in `xorq_buckaroo.py`), so no polars import on the xorq path. Locked it in with a subprocess regression test in 9fb75295 — `tests/unit/test_xorq_polars_isolation.py` runs each side's top-level import with a meta-path blocker for the other extra, so this can't silently regress.

buckaroo/xorq_buckaroo.py
```diff
@@ -0,0 +1,367 @@
...
+                orig_col = processed_sd[sort]['orig_col_name']
+                slice_df = self._sliced_pandas(
+                    processed_df, start, end, sort_col=orig_col, ascending=(sort_dir == 'asc'))
+                self.send(
+                    {"type": "infinite_resp", 'key': new_payload_args,
+                     'data': [], 'length': total_length},
+                    [to_parquet(slice_df)])
+                return
+
+            slice_df = self._sliced_pandas(processed_df, start, end)
```
paddymul: It is unclear to me where the slicing occurs.  I don't see an `execute()` call. The expression should stay lazy until this point. and it should only be executed over the small portion that is being displayed.

I realize that summary stats need to call execute, but that is for aggregates, for the core expression, it shouldn't receive a bare execute call that isn't limitted by ranges.

paddymul: Addressed in 98d5bf06. Two changes:

1. Inlined the window query into `_handle_payload_args`. Renamed the helper to `_execute_window` and rewrote its docstring to spell out that `expr.limit(end - start, offset=start).execute()` is the *only* place the expression materialises and only over rows `[start, end)`. The expression stays lazy until that line.

2. The unbounded fallback `df_or_expr.execute()` in `XorqSampling.serialize_sample` now raises `RuntimeError` instead. It was unreachable today (basic widget uses `serialize_limit=5_000`; infinite widget bypasses `serialize_sample` via `skip_main_serial=True`) but it shouldn't have been there — exactly the kind of thing your note was warning against.

Issue #706, opened 2026-05-07
Pandas widget construction: traitlets equality check on full DataFrame eats hundreds of ms


paddymul:
## Summary

Constructing `BuckarooInfiniteWidget` or `BuckarooWidget` on a moderately large pandas DataFrame is dominated by a `DataFrame.__eq__` triggered by `traitlets.set` change-detection. On `Boston restaurant inspections` (883k × 26, mostly object dtype), this is **586 ms out of 755 ms total** (78% of widget construction).

The stats pipeline itself is only ~140 ms — it's not the bottleneck.

## Repro

```bash
.venv/bin/python scripts/perf/perf_widgets.py
.venv/bin/python scripts/perf/perf_profile.py
```

(Both scripts are on the `perf/polars-stats-instrumentation` branch.)

## Profile (BuckarooInfiniteWidget @ 883k × 26 pandas, after warmup)

```
ncalls  tottime  function
     2    0.586  array_ops.py:113(comp_method_OBJECT_ARRAY)
    46    0.042  algorithms.py:963(value_counts_arraylike)
     1    0.143  dataflow.py:199(_summary_sd)
```

`print_callers` chain:

```
comp_method_OBJECT_ARRAY  <- comparison_op
                          <- frame.py:7894 (DataFrame._cmp_method)
                          <- arraylike.py:38 (DataFrame.__eq__)
                          <- common.py:62 (new_method)
                          <- traitlets.py:689 (set)     ← 2 calls, 586 ms
```

Two assignments to DataFrame-bearing `Any()` traits during construction. Each one performs a full element-wise `==` between the old and new value to decide whether to fire `@observe` callbacks.

Scales linearly with cells:

| dataset | cells | comp_method_OBJECT_ARRAY (2 calls) |
| --- | --- | --- |
| synth 100k × 8 | 800k | 13 ms |
| Boston 883k × 26 | 23M | 586 ms |

## Why polars dodges it

Polars `df == df2` returns a polars DataFrame. Traitlets's reduction to `bool(...)` for the equality check either raises or short-circuits without iterating elements, so polars widgets never pay this cost — `PolarsBuckarooInfiniteWidget` finishes in 96 ms on the same data.

## Affected traits

`buckaroo/dataflow/dataflow.py` declares ~10 `Any()` traits that may hold DataFrames:

- `raw_df`, `sampled_df`, `cleaned`, `processed_result`
- `summary_sd`, `merged_sd`
- `widget_args_tuple`
[...]

PR #710, opened 2026-05-08
fix(xorq): #709 cut SQL queries 51→9 on widget construction

### LINKED ISSUE #709: XorqStatPipeline issues 51 SQL queries per widget construction (should be 1-2)
## Summary

Constructing `XorqBuckarooWidget` issues **51 separate SQL queries** to the backend for an 8-column dataframe (50k rows), and **87 queries** for a 26-column dataframe. This is the dominant cost of widget construction (700–800 ms vs polars's 60–100 ms doing the same work in-memory).

The expectation, based on how polars does it, is that the entire stats pipeline collapses into **one** complex SQL query (or at most a small constant number).

## Repro

```bash
.venv/bin/python -W ignore::DeprecationWarning -c "
import warnings
warnings.filterwarnings('ignore')
import pandas as pd
import xorq.api as xo
from buckaroo.pluggable_analysis_framework.xorq_stat_pipeline import XorqStatPipeline
from buckaroo.xorq_buckaroo import XorqBuckarooWidget
import sys, traceback
sys.path.insert(0, 'scripts/perf')
from perf_data import make_pandas
from collections import Counter

orig = XorqStatPipeline._execute
sites = []
def counting(self, q):
    stack = traceback.extract_stack()
    site = next((f'{f.filename.split(\"/\")[-1]}:{f.lineno}:{f.name}'
                 for f in reversed(stack)
                 if 'buckaroo' in f.filename and 'xorq_stat_pipeline' not in f.filename), '?')
    sites.append(site)
    return orig(self, q)
XorqStatPipeline._execute = counting

pdf = make_pandas(50_000)
con = xo.connect()
table = con.create_table('t', pdf)
XorqBuckarooWidget(table)
print(f'total queries: {len(sites)}')
for site, n in Counter(sites).most_common():
    print(f'  {n:3}  {site}')
"
```

## Output (datafusion backend, 50k × 8)

```
total queries: 51
   40  xorq_stats_v2.py:245:_categorical_histogram
    6  xorq_stats_v2.py:213:_numeric_histogram
    4  dataflow.py:421:_get_summary_sd          # the batched-aggregate query
    1  dataflow.py:345:setup_options_from_analysis
```

Same numbers on duckdb backend.

On 50k × 26 (Boston restaurant inspections head): **87 queries**.

### A. Pipeline runs the full query set 4× per widget construction

`dataflow.py:421:_get_summary_sd` is called **4 times**. Each call re-runs the entire stat pipeline — meaning 4 batched aggregates and ~4× the per-column histogram queries. Likely a traitlet observer chain firing on every transient trait assignment during `__init__`. Most of these calls are redundant; the final `widget_args_tuple` is the only one whose output is consumed.

A user constructing the widget once should be paying for one pipeline run, not four. This is similar in spirit to issue #706 (over-eager change-detection on traitlets-held DataFrames) — except here it's doing actual work, not just comparing.

### B. Histograms are one query per column instead of one query total

`_numeric_histogram` and `_categorical_histogram` (`buckaroo/customizations/xorq_stats_v2.py:191` and `:238`) each issue a per-column query — an `expr.group_by(col).aggregate(count).order_by(...).limit(10)` for categoricals, a bucketed group-by for numerics. **40 + 6 = 46 of the 51 queries (90%) are histograms.**

For categorical histograms specifically, all N column queries can be expressed as one SQL pass using `array_agg`/`struct_agg` over per-column top-K subqueries, or one CTE that pivots. For numerics, the bucket aggregation can be unified across columns by using the already-known `min`/`max` per column (which the batched aggregate already computes).

The Phase-1 batched aggregate (`table.aggregate(...)` for length / null_count / min / max / etc.) already proves the pattern works — extend it to cover histograms too.

## Why this matters

Comparison to pandas/polars on the same data (no SQL, all in-memory):

| widget | Boston 883k × 26 |
| --- | --- |
| `BuckarooInfiniteWidget` (pandas, post-#706 fix) | 715 ms |
| `PolarsBuckarooInfiniteWidget` | 96 ms |
| `XorqBuckarooInfiniteWidget` (datafusion) | 732 ms |
| `XorqBuckarooInfiniteWidget` (duckdb) | 680 ms |

Even on an in-process embedded engine like datafusion, per-query overhead × dozens of queries ≈ pandas's worst case (which itself is dominated by an unrelated traitlets bug). On a remote backend (Trino, DuckDB-over-network, etc.) the per-query latency would push this from "slow" into "completely unusable."

## Suggested fix shape

In rough priority order:

1. **Stop redundant pipeline runs.** Investigate why `_get_summary_sd` fires 4× per construction — likely some `@observe` chain. Coalesce into one run, or short-circuit when `widget_args_tuple` hasn't actually changed shape.
2. **Batch the histograms.** Either:
   - Fold all per-column histogram aggregates into the existing Phase-1 `table.aggregate(...)` call (one SELECT, many `…name(f"{col}|histogram_bucket_{i}")` columns).
   - Or run all categorical histograms as one `UNPIVOT` / `UNION ALL` query, all numeric histograms as another.
3. **Goal: one batched query for stats, plus one per-column-group histogram query** (still much better than 46). Long-term: one query for everything, like polars.

## Test plan

- [ ] Add a regression test that asserts `len(queries) <= N` for some small N (probably ≤ 10) on a 26-column input. Pipeline-runs-4× alone should be caught with `≤ 2`.
- [ ] Re-run `scripts/perf/perf_xorq.py`. Boston datafusion target: well under 200 ms (down from 793 ms).
- [ ] Confirm the resulting summary stats are byte-identical to the pre-fix pipeline on a representative test fixture.

## References

- Investigation: `docs/research/perf-polars-stats.md` (xorq section)
- Bench: `scripts/perf/perf_xorq.py` (on branch `perf/polars-stats-instrumentation`)
- Stat pipeline: `buckaroo/pluggable_analysis_framework/xorq_stat_pipeline.py:168` (`process_table`)
- Histograms: `buckaroo/customizations/xorq_stats_v2.py:191,238`

paddymul:
## Summary

Closes #709. Three independent fixes that drop XorqBuckarooWidget construction SQL query count from 51 → 9 (8-col synthetic) and 87 → 27 (26-col Boston restaurants), with proportional perf wins (2.3-3.9× faster widget construction).

## What's wrong (today)

`XorqBuckarooWidget(expr)` issues 51 separate SQL queries during construction on an 8-column input. Three stacked causes:

1. **`_summary_sd` observer fires twice.** `DataFlow._operation_result` re-sets `self.operations = result[3]`, which triggers another `_operation_result` cascade because the observer also watches `operations`. Every cascade rebuilds `processed_result` as a fresh tuple wrapper, so `_summary_sd`'s observer fires again even though `processed_df` identity hasn't changed.
2. **`XorqDfStatsV2` constructs `XorqStatPipeline(unit_test=True)`** (the default) on every widget. The unit test runs the full pipeline against `PERVERSE_DF` — a SQL roundtrip per stat per col. `PlDfStatsV2` already passes `unit_test=False` for the same reason; xorq missed it.
3. **`verify_analysis_objects` also defaults to `unit_test=True`.** Called from `setup_options_from_analysis` on every widget.

## Fixes

`buckaroo/dataflow/dataflow.py:_summary_sd` — dedupe by `(id(processed_df), id(analysis_klasses))`:

```python
_summary_sd_cache_key = (None, None)

@observe('processed_result', 'analysis_klasses')
@exception_protect('summary_sd-protector')
def _summary_sd(self, change):
    df = self.processed_df
    klasses = self.analysis_klasses
    if (id(df), id(klasses)) == self._summary_sd_cache_key:
        return
    self._summary_sd_cache_key = (id(df), id(klasses))
    result_summary_sd, errs = self._get_summary_sd(df)
    self.summary_sd = result_summary_sd
    self.errs = errs
```

Identity comparison is right here — the `==` would be cheap in xorq (returns an expression) but using `is`/`id` is consistent with `DfTrait` (#706).

`buckaroo/pluggable_analysis_framework/xorq_stat_pipeline.py:XorqDfStatsV2.__init__` and `verify_analysis_objects` — pass `unit_test=False`. DAG validation (the part that actually catches misconfig) still runs in `XorqStatPipeline.__init__`; only the per-widget `PERVERSE_DF` SQL roundtrip is skipped.

## Numbers

Query count after fix:

| dataset | pre | post |
| --- | --- | --- |
| `_expr()` 3 cols | 41 | **4** |
| synth 50k × 8 | 51 | **9** |
| Boston 50k × 26 | 87 | **27** |

Widget construction time (warm best of 3, after `create_table`):

| dataset | xorq[df] PRE | xorq[df] POST | speedup |
| --- | --- | --- | --- |
| synth 100k × 8 | 320 ms | **81 ms** | **3.9×** |
| synth 500k × 8 | 437 ms | **133 ms** | 3.3× |
| Fielding 174k × 18 | 555 ms | **209 ms** | 2.7× |
| Boston 883k × 26 | 793 ms | **318 ms** | 2.5× |

| dataset | xorq[duckdb] PRE | xorq[duckdb] POST | speedup |
| --- | --- | --- | --- |
| synth 100k × 8 | 364 ms | **103 ms** | 3.5× |
| synth 500k × 8 | 396 ms | **118 ms** | 3.4× |
| Fielding 174k × 18 | 690 ms | **269 ms** | 2.6× |
| Boston 883k × 26 | 744 ms | **325 ms** | 2.3× |

xorq[datafusion] is now competitive with polars on small-N (synth 100k: 81 vs 99 ms) and beats pandas on large-N (Boston: 318 vs 758 ms post-#706 fix).

## What's still left

The remaining ~9 queries on 8 cols are 1 batched aggregate + 1 histogram per column. Folding histograms into the batched aggregate would drop the count to 1-2 regardless of column count — that's the polars-equivalent end state. Out of scope for this PR; deserves its own design (numeric histograms can use the already-computed min/max via a 2-pass batch; categorical top-K is harder to batch in ibis).

## Test plan

- [x] New `TestQueryCount.test_widget_construction_query_count` asserts ≤ 6 queries for the 3-col fixture (was 41).
- [x] Full unit suite stays green (956 passed).
- [x] Re-run `scripts/perf/perf_xorq.py` — numbers above match the report.

## Commits

1. **`test(xorq): failing query-count regression test for #709`** — pushed first; CI fails on this commit (assertion: 41 ≤ 6).
2. **(coming next push)** the fix.

### DIFF
--- a/buckaroo/dataflow/dataflow.py
+++ b/buckaroo/dataflow/dataflow.py
@@ -222,10 +222,23 @@ class DataFlow(ABCDataflow):
             ret_summary[col] = {}
         return ret_summary, {}
 
+    _summary_sd_cache_key = (None, None)
+
     @observe('processed_result', 'analysis_klasses')
     @exception_protect('summary_sd-protector')
     def _summary_sd(self, change):
-        result_summary_sd, errs  = self._get_summary_sd(self.processed_df)
+        # Dedupe: the autocleaning operations cascade re-fires
+        # _operation_result → cleaned → processed_result with a freshly-built
+        # tuple wrapper, which makes this observer fire twice per widget
+        # construction even when processed_df identity is unchanged.
+        # Skip when neither the dataframe nor analysis_klasses has actually
+        # changed since the last run. See issue #709.
+        df = self.processed_df
+        klasses = self.analysis_klasses
+        if (id(df), id(klasses)) == self._summary_sd_cache_key:
+            return
+        self._summary_sd_cache_key = (id(df), id(klasses))
+        result_summary_sd, errs  = self._get_summary_sd(df)
         self.summary_sd = result_summary_sd
         self.errs = errs
 
--- a/buckaroo/pluggable_analysis_framework/xorq_stat_pipeline.py
+++ b/buckaroo/pluggable_analysis_framework/xorq_stat_pipeline.py
@@ -342,11 +342,17 @@ class XorqDfStatsV2:
 
     @classmethod
     def verify_analysis_objects(cls, objs):
-        XorqStatPipeline(objs)
+        # unit_test=False to skip the per-widget PERVERSE_DF pipeline run
+        # (issue #709). DAG validation still runs as part of __init__.
+        XorqStatPipeline(objs, unit_test=False)
 
     def __init__(self, table, col_analysis_objs, operating_df_name=None, debug=False):
         self.table = table
-        self.ap = XorqStatPipeline(col_analysis_objs)
+        # Skip the unit_test PERVERSE_DF run on each widget construction —
+        # it doubles the SQL query count (issue #709). The DAG-validation
+        # cost is already paid by verify_analysis_objects on first set up
+        # and by the test suite. Mirrors PlDfStatsV2.
+        self.ap = XorqStatPipeline(col_analysis_objs, unit_test=False)
         self.operating_df_name = operating_df_name
         self.debug = debug
         self.sdf, self.errs = self.ap.process_table_v1_compat(self.table)
--- a/docs/research/perf-polars-stats.md
+++ b/docs/research/perf-polars-stats.md
@@ -390,16 +390,27 @@ Infinite variant on the same data:
    number** in any of the harnesses — pandas DataFrame → duckdb table
    ingestion of a 376 MB string-heavy frame.
 
-### What's next for xorq perf
+### Update — #709 fix landed (PR #710)
 
-Not investigated here, but worth considering:
-- Profile `XorqStatPipeline.process_df` to see whether the cost is
-  per-query overhead (network/IPC) or actual SQL execution. If it's
-  the former, batching multiple stats into a single SQL query (one
-  SELECT with many aggregations) would collapse the round-trip count.
-- The `XorqDfStatsV2` likely benefits from the #706 traitlets fix
-  (it inherits the DataFlow traits), but the win is small there
-  because xorq exprs don't have an O(rows × cols) `__eq__`.
+Profiling confirmed it's per-query overhead, not SQL execution. Three
+fixes (`_summary_sd` dedupe, `XorqDfStatsV2` and
+`verify_analysis_objects` both pass `unit_test=False`) cut the query
+count by ~80% and the wall time by 2.3-3.9×:
+
+| dataset | xorq[df] PRE | xorq[df] POST | xorq[duckdb] PRE | xorq[duckdb] POST |
+| --- | --- | --- | --- | --- |
+| synth 100k × 8 | 320 ms | **81 ms** (3.9×) | 364 ms | **103 ms** (3.5×) |
+| synth 500k × 8 | 437 ms | **133 ms** (3.3×) | 396 ms | **118 ms** (3.4×) |
+| Fielding 174k × 18 | 555 ms | **209 ms** (2.7×) | 690 ms | **269 ms** (2.6×) |
+| Boston 883k × 26 | 793 ms | **318 ms** (2.5×) | 744 ms | **325 ms** (2.3×) |
+
+Post-fix query count: 4 (3-col fixture), 9 (8 cols), 27 (26 cols).
+Remaining is 1 batched aggregate + 1 histogram per column.
+
+Still TODO: fold histograms into the batched aggregate to drop the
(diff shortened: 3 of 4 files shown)


paddymul:
Superseded by stacked PRs: monitoring → #711, xorq (#709) → #714. The original combined branch is preserved for history but won't be merged.

PR #721, opened 2026-05-12
build: make buckaroo-js-core publishable to npm + built-pkg integration test

paddymul:
## Summary

Two related changes:

### 1. Make `buckaroo-js-core` publishable to npm

`packages/buckaroo-js-core/package.json` — added description, keywords, license, author, repository, homepage, bugs; added `module` + `types` alongside the existing `main`; added an `exports` map (`.`, `./style.css`, `./dist/style.css`, `./package.json`); set `publishConfig.access = "public"`; added `prepublishOnly = "pnpm run build"`; rewrote `files`. Added `LICENSE` (copied from repo root, BSD-3-Clause) and replaced the Vite template `README.md` with one useful for npm consumers.

### 2. Integration test that renders the built package

`integration-tests/built-pkg/` — a tiny tree at the repo root (outside `packages/`, so outside the pnpm workspace) that installs the locally-packed `buckaroo-js-core` tarball via **plain `npm install`** (not pnpm) and renders a `DFViewer` with three sample rows. Playwright asserts the cell values are in the DOM **and** that no fetched URL contains `packages/buckaroo-js-core/src` — definitive proof the rendered code came from the tarball, not workspace source.

Run it: `bash integration-tests/built-pkg/run.sh`

## Anywidget impact

None by design. `packages/js/{widget,standalone,static-embed}.tsx` resolve `buckaroo-js-core` via bare specifier and a relative CSS path — both still land on the same `dist/index.esm.js` (`main` is unchanged; the new `exports["."]` aliases to the same file).

## Test plan

- [x] `pnpm --filter buckaroo-js-core run build` — identical output (1504 modules)
- [x] `pnpm --filter buckaroo-widget run build` / `build:standalone` / `build:static` — all green
- [x] `pnpm --filter buckaroo-js-core test` — 96/96 jest
- [x] `pnpm pack` — 5.0 MB tarball with LICENSE, README, dist/, package.json
- [x] `bash integration-tests/built-pkg/run.sh` — pack + plain-npm install + Playwright, 2/2 pass

## Out of scope

- Actually publishing to npm (decide version + access after merge).
- Trimming Storybook / `*.test.d.ts` files from the tarball — minor cosmetic noise; leaving alone to keep this PR mechanical.
- Wiring the integration test into CI — currently a manual pre-publish gate; can be added to the release workflow later.


paddymul:
Addressed in 5558fc0f. Dropped the `require` condition rather than adding a real CJS build — `"type": "module"` plus the `.js` extension on the cjs output made the advertised `require` path either throw (Node 18) or return empty (Node 20+). The package is ESM-only in practice and all consumers (workspace anywidget + the new built-pkg integration test) use ESM. Real CJS support can be a follow-up if a CJS consumer surfaces; it'd need a `.cjs` rename in the vite config.

PR #722, opened 2026-05-12
test(integration): verify buckaroo-js-core renders from the built tarball

paddymul:
## Summary

New tree at `integration-tests/built-pkg/` that installs the locally-packed `buckaroo-js-core` tarball into its own `node_modules` (via **plain `npm install`**, outside the pnpm workspace) and renders a `DFViewer` with three sample rows. Playwright asserts the cell values are in the DOM.

The point is to catch any way the published npm package could break — missing `main`/`module`/`types`, mis-routed `exports`, stripped `dist/` files, peer-dep regressions, CSS subpath failure outside the workspace — that the in-workspace tests don't see because they always read from source.

## Why outside `packages/`

The Buckaroo pnpm workspace is at `packages/pnpm-workspace.yaml`. If this test lived under `packages/`, pnpm would symlink `buckaroo-js-core` back to source and we'd silently degrade to testing workspace. Two safeguards:

1. **Outside the workspace** — `integration-tests/` is not a workspace member.
2. **Plain `npm install`** in `run.sh`, not pnpm.
3. Plus a Playwright check that asserts no fetched resource URL contains `packages/buckaroo-js-core/src` during render.

## Files (10)

```
integration-tests/built-pkg/
├── package.json           # buckaroo-js-core = file:./buckaroo-js-core.tgz
├── vite.config.ts         # pinned to 127.0.0.1:5174 for Playwright reach
├── tsconfig.json
├── index.html
├── src/main.tsx           # imports { DFViewer } from "buckaroo-js-core"
├── playwright.config.ts   # webServer = npm run dev
├── pw-tests/renders.spec.ts
├── run.sh                 # one-shot orchestrator
├── .gitignore             # excludes node_modules, lockfile, tarball, reports
└── README.md
```

`node_modules/`, `package-lock.json`, `buckaroo-js-core*.tgz`, and `dist/` are all gitignored — regenerated by `run.sh` on every run.

## Test plan

- [x] `bash integration-tests/built-pkg/run.sh` from a clean state — 2/2 Playwright tests pass.
- [x] First-time IPv4/IPv6 surprise (Vite was on `::1`, Playwright on `127.0.0.1`) fixed by pinning `vite.config.ts` `server.host`.
- [x] No changes to `packages/buckaroo-js-core/` or `packages/js/`.

## When to run

Manually before publishing to npm; after any change to `packages/buckaroo-js-core/package.json`'s exports/files/main/module fields; after react/ag-grid/vite bumps. Not yet wired into CI — could be added to the release workflow as a pre-publish gate.

## Relationship to #721

This PR is independent and works against `main` today (uses `buckaroo-js-core/dist/style.css` for the CSS subpath, which works without the `exports` field). It continues to work after #721 lands.


paddymul:
Folded into #721 as requested.

PR #744, opened 2026-05-15
feat(jlisp): transform may return (df, sd_updates) tuple

paddymul:
## Summary

Extends the jlisp interpreter contract so a Command's `transform` can return either a bare df or a 2-tuple `(df, sd_updates)`. The optional second element is a partial `SDType` (`{col: {key: value}}`) that gets merged into `cleaning_sd` after the operation runs. This is the foundation that lets a Command thread styling-relevant metadata (e.g. a search term that becomes a highlight phrase downstream) without having to round-trip it through the dataframe.

## Design

- `_wrap_transform` in `buckaroo/jlisp/configure_utils.py` inspects the return value: if it's a 2-tuple `(df, dict)` it accumulates the dict into a per-call sd_accumulator and returns the bare df to the rest of the interpreter (so existing call sites stay unchanged).
- After `buckaroo_transform` runs, the accumulator is readable via `buckaroo_transform.get_last_sd_updates()`.
- `sd_accumulator.clear()` runs at the start of each `buckaroo_transform` call so updates don't leak across invocations.

No existing Command is updated in this PR — this is the pure contract change. The first consumer (polars Search → highlight) lands in the follow-up.

## Test plan

- [x] `pytest tests/unit/jlisp/ tests/unit/dataflow/` — 134 pass
- [ ] CI green

### DIFF
--- a/buckaroo/customizations/polars_commands.py
+++ b/buckaroo/customizations/polars_commands.py
@@ -6,12 +6,19 @@ from ..jlisp.lisp_utils import s
 #from ..auto_clean.cleaning_commands import (to_bool, to_datetime, to_int, to_float, to_string)
 
 class Command(object):
-    @staticmethod 
+    """A Command's `transform` returns either the new df, or a 2-tuple
+    `(df, sd_updates)` where sd_updates is a partial SDType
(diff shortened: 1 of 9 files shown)


paddymul:
superseded by #755

PR #745, opened 2026-05-15
feat(polars-search): plumb search term to JS as highlight_regex

paddymul:
## Summary

Polars `Search.transform` now returns `(filtered_df, sd_updates)` so the search term flows into `cleaning_sd` as `highlight_regex` on every polars-String column. `DefaultMainStyling.style_column` reads `highlight_phrase` / `highlight_regex` / `highlight_color` off `col_meta` and threads them into the string `displayer_args`, where the JS side already renders matches as `<mark>`.

This is the first concrete consumer of the `(df, sd_updates)` 2-tuple transform contract from **#744**.

## Depends on

**#744** — the jlisp interpreter contract change must merge first. CI here will fail until that lands; after it does, this PR will rebase and the duplicate commit drops via patch-id.

## Pieces

- **`feat(jlisp)`** — temporarily included; drops on rebase after #744.
- **`fix: rename op-contributed sd keys + harden style_column against missing _type`** — autocleaning rewrites the op-supplied keys via `old_col_new_col(cleaned_df)` so they line up with buckaroo's internal a/b/c names; `style_column` falls back to obj when `_type` is absent.
- **`feat(polars-search): plumb search term to JS highlight via sd`** — the Search transform change + styling-layer extraction.
- **`test(polars): end-to-end highlight delivery into df_viewer_config`** — integration test asserting the full pipeline lands `highlight_regex` on the correct column in `df_viewer_config`.

## Test plan

- [x] `pytest tests/unit/dataflow/ tests/unit/commands/polars_command_test.py` — 117 pass
- [ ] CI green (after #744 merges + rebase)

### DIFF
--- a/buckaroo/customizations/polars_commands.py
+++ b/buckaroo/customizations/polars_commands.py
@@ -6,12 +6,19 @@ from ..jlisp.lisp_utils import s
 #from ..auto_clean.cleaning_commands import (to_bool, to_datetime, to_int, to_float, to_string)
 
 class Command(object):
-    @staticmethod 
+    """A Command's `transform` returns either the new df, or a 2-tuple
+    `(df, sd_updates)` where sd_updates is a partial SDType
+    ({col: {key: value}}) to be merged into cleaning_sd. The 2-tuple form
+    lets a Command thread styling-relevant metadata (e.g. a search term
+    that becomes a highlight phrase downstream) without round-tripping
+    through the df."""
+
+    @staticmethod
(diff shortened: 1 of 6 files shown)


paddymul:
Superseded by #758, which is the same work rebased onto main after #755 merged. The obsolete tuple-contract commit (`feat(jlisp): transform may return (df, sd_updates) tuple`) is dropped; `Search.transform` now returns `SDResult` from #755 instead.

PR #760, opened 2026-05-17
feat(js-core): split BuckarooView from BuckarooServerView for injectable IModel (#759)

### LINKED ISSUE #759: BuckarooServerView should accept an injectable model (IModel) so hosts can keep the WebSocket out of the webview
## Problem

`BuckarooServerView` (added in #724) only accepts a `wsUrl: string` prop and constructs its own `WebSocketModel` internally:

```tsx
// packages/buckaroo-js-core/src/server/BuckarooServerView.tsx
export interface BuckarooServerViewProps {
    wsUrl: string;
    renderConnecting?: () => React.ReactNode;
    renderError?: (err: Error) => React.ReactNode;
    onMetadata?: (metadata: BuckarooServerMetadata, prompt?: string) => void;
    style?: React.CSSProperties;
    className?: string;
}
```

This means a host app that mounts `<BuckarooServerView>` necessarily **opens a WebSocket from its renderer / webview** to the buckaroo sidecar. For Tauri / Electron / Wails desktop hosts that explicitly avoid renderer-side network sockets (CSP-tightening, sandbox compatibility, no-listening-port property), this is a non-starter — the whole reason `buckaroo-tauri` exists is to relay the buckaroo WS protocol through the Rust supervisor over Tauri IPC so the webview never opens a socket itself.

Today, integrating `BuckarooServerView` into a `buckaroo-tauri`-using host means either:

1. **Sacrifice the no-WS property** — widen the host's CSP `connect-src` to `ws://127.0.0.1:*`, accept that the renderer talks to the local sidecar directly. (This is what my xorq-desktop spike currently does.)
2. **Don't use `BuckarooServerView`** — assemble the inner widget tree manually with a `TauriIPCModel`. Doable, but the host now owns logic (row cache wiring, mode dispatch, pre-resolution) that `BuckarooServerView` was specifically built to encapsulate.

Both are bad trades for what could be a single optional prop.

## Precedent: nteract

Modern nteract (`nteract/desktop`, Tauri-based — [Cargo workspace](https://github.com/nteract/desktop/blob/main/Cargo.toml)) **categorically does not open WebSockets from the renderer**. Their architecture:

```
webview ──Tauri IPC── notebook crate (Rust) ──UnixStream── runtimed daemon ──ZMQ── kernel
```

`WebSocket` only appears in their codebase as (a) a comment about a future web-app target in [`packages/runtimed/src/transport.ts`](https://github.com/nteract/desktop/blob/main/packages/runtimed/src/transport.ts), and (b) a dev-only Vite relay (`apps/notebook/vite-plugin-browser-relay.ts`). All production kernel comms go renderer → IPC → Rust → daemon → ZMQ → kernel. The biggest Tauri-Jupyter shop deliberately took the IPC-relay route over a renderer-side WS.

`buckaroo-tauri` is the same shape, just for buckaroo's server protocol instead of Jupyter's: the Rust supervisor opens an internal `127.0.0.1:N` WebSocket to the Python sidecar and relays messages to the webview via Tauri events. `buckaroo-tauri-adapter` already exports `TauriIPCModel` (implements `IModel`) for exactly this use case — but `BuckarooServerView` has no way to use it today.

## Proposed change

Make the transport injectable. I see two clean shapes; happy with either, lean toward #2.

### Option 1 — optional `model` prop, mutually exclusive with `wsUrl`

```tsx
export type BuckarooServerViewProps =
  | { wsUrl: string;  model?: never; /* other props */ }
  | { wsUrl?: never;  model: IModel; mode: BuckarooServerMode; /* other props */ };
```

When `model` is provided, skip the `WebSocketModel` construction and use it as-is. Caller is then responsible for ensuring `initial_state` has been received (since there's no `connecting` phase). `mode` would need to be passed explicitly since it's no longer derived from a server message the component itself sees.

Drawback: prop shape gets pleonastic, the `initial_state` lifecycle for the injected case is awkward.

### Option 2 — split into transport-agnostic inner + WS-specific wrapper (preferred)

```tsx
// New, transport-agnostic. Takes an already-connected model + the initial_state it produced.
export interface BuckarooViewProps {
    model: IModel;
    initialState: Record<string, unknown>;
    mode: BuckarooServerMode;
    /* renderConnecting unnecessary — caller handles pre-connection */
    renderError?: (err: Error) => React.ReactNode;
    onMetadata?: (metadata: BuckarooServerMetadata, prompt?: string) => void;
    style?: React.CSSProperties;
    className?: string;
}
export function BuckarooView(props: BuckarooViewProps): JSX.Element { /* the existing widget-dispatch + row-cache + pre-resolution logic */ }

// Existing, unchanged public surface. Thin wrapper that constructs WebSocketModel,
// waits for initial_state, then renders <BuckarooView>.
export function BuckarooServerView(props: BuckarooServerViewProps): JSX.Element { /* … */ }
```

This:
- Preserves the existing `BuckarooServerView` API surface byte-for-byte. No breaking changes.
- Makes the model-injection case a first-class export: `import { BuckarooView } from "buckaroo-js-core"` and pass `TauriIPCModel` + a pre-collected `initial_state` (the adapter's `waitForInitialState()` already exists for this exact purpose).
- Has a clear separation of concerns: connection handling lives in `BuckarooServerView`, rendering lives in `BuckarooView`.
- Easier to test — `BuckarooView` can be exercised with a mock IModel without any WS plumbing.

## Use case (concrete)

In a xorq-desktop spike I'm running today, my mount looks like:

```tsx
<BuckarooServerView
  wsUrl={buckarooWsUrl(`http://127.0.0.1:${sidecarPort}`, sessionId)}
  onMetadata={(m) => console.log(m.path)}
/>
```

With option 2 it would become:

```tsx
const initial = await waitForInitialState(); // adapter helper, already exists
const model = new TauriIPCModel(initial);
<BuckarooView
  model={model}
  initialState={initial}
  mode={initial.mode as BuckarooServerMode}
  onMetadata={(m) => console.log(m.path)}
/>
```

No webview WS, CSP can stay tight (`connect-src ipc: http://ipc.localhost`), and `buckaroo-tauri`'s "renderer never opens a socket" property holds.

## Out of scope

- Whether `buckaroo-tauri-adapter`'s `TauriIPCModel` needs API tweaks to match the upstream `IModel` precisely — that can be a follow-up. Today it's vendored.
- Renaming / deprecation. `BuckarooServerView` keeps its current API and behavior.

Happy to draft the PR if direction is acceptable.

paddymul:
## Summary

- Extract a transport-agnostic \`BuckarooView\` from \`BuckarooServerView\`. Existing \`BuckarooServerView\` API is unchanged — it's now a thin wrapper that opens the WebSocket, awaits \`initial_state\`, and delegates rendering to \`BuckarooView\`.
- Add an \`IModel\` interface — the minimal model surface (\`send\`/\`get\`/\`set\`/\`save_changes\`/\`on\`/\`off\`) that \`BuckarooView\` consumes. \`WebSocketModel\` is structurally assignable to it; host adapters like \`buckaroo-tauri-adapter\`'s \`TauriIPCModel\` can now drive \`BuckarooView\` directly without opening a renderer-side WebSocket.

Closes #759.

## Why

For Tauri / Electron / Wails hosts that explicitly avoid renderer-side network sockets (CSP tightening, sandbox compat, no-listening-port property), \`BuckarooServerView\` was a non-starter — it constructs its own \`WebSocketModel\` internally. The two prior options were both bad trades: widen CSP \`connect-src\` to \`ws://\`, or rebuild the row-cache / mode-dispatch / pre-resolution wiring by hand. This PR makes the wiring a first-class export so host adapters can inject a pre-connected model.

## API

\`\`\`tsx
import { BuckarooView, IModel } from \"buckaroo-js-core\";

const initial = await waitForInitialState();      // adapter helper
const model: IModel = new TauriIPCModel(initial); // your IPC-backed adapter

<BuckarooView
  model={model}
  initialState={initial}
  mode={initial.mode}                             // \"viewer\" | \"buckaroo\"
  onMetadata={(m) => console.log(m.path)}
/>
\`\`\`

No webview WS, CSP can stay tight, and \`buckaroo-tauri\`'s \"renderer never opens a socket\" property holds.

## Backwards compatibility

- \`BuckarooServerView\` API surface unchanged: same props (\`wsUrl\`, \`renderConnecting\`, \`renderError\`, \`onMetadata\`, \`style\`, \`className\`), same connect → wait-for-initial_state → render flow.
- Re-exports \`BuckarooServerMode\` and \`BuckarooServerMetadata\` from the same path they shipped at, so existing imports keep resolving.

## Test plan

- [x] \`pnpm test\` — 235 tests pass (includes new \`BuckarooView.test.tsx\` covering the IModel injection path with a fake model)
- [x] \`pnpm tsc --noEmit\` clean
- [x] \`pnpm run build\` clean

### DIFF
--- a/packages/buckaroo-js-core/src/index.ts
+++ b/packages/buckaroo-js-core/src/index.ts
@@ -16,6 +16,7 @@ import { parquetRead, parquetMetadata } from 'hyparquet';
 import { resolveDFData, resolveDFDataAsync, preResolveDFDataDict } from './components/DFViewerParts/resolveDFData';
 import { BuckarooStaticTable } from './components/BuckarooStaticTable';
 import { BuckarooServerView, buckarooWsUrl } from './server/BuckarooServerView';
+import { BuckarooView } from './server/BuckarooView';
 import { WebSocketModel } from './server/WebSocketModel';
 
 import { HistogramCell } from "./components/DFViewerParts/HistogramCell";
@@ -57,6 +58,7 @@ export default {
     preResolveDFDataDict,
     BuckarooStaticTable,
     BuckarooServerView,
+    BuckarooView,
     buckarooWsUrl,
     WebSocketModel,
 };
@@ -86,6 +88,11 @@ export {
     preResolveDFDataDict,
     BuckarooStaticTable,
     BuckarooServerView,
+    BuckarooView,
     buckarooWsUrl,
     WebSocketModel,
 };
+
+export type { IModel } from './server/IModel';
+export type { BuckarooViewProps, BuckarooServerMode, BuckarooServerMetadata } from './server/BuckarooView';
+export type { BuckarooServerViewProps } from './server/BuckarooServerView';
--- a/packages/buckaroo-js-core/src/server/BuckarooServerView.tsx
+++ b/packages/buckaroo-js-core/src/server/BuckarooServerView.tsx
@@ -1,36 +1,17 @@
 import * as React from "react";
 
-import { BuckarooInfiniteWidget, DFViewerInfiniteDS, getKeySmartRowCache } from "../components/BuckarooWidgetInfinite";
 import { preResolveDFDataDict } from "../components/DFViewerParts/resolveDFData";
 import { WebSocketModel } from "./WebSocketModel";
-import { DFMeta, BuckarooState, BuckarooOptions } from "../components/WidgetTypes";
-import { CommandConfigT } from "../components/CommandUtils";
-import { Operation } from "../components/OperationUtils";
-import { OperationResult, baseOperationResults } from "../components/DependentTabs";
-import { DFData, DFDataOrPayload } from "../components/DFViewerParts/DFWhole";
-import { IDisplayArgs } from "../components/DFViewerParts/gridUtils";
+import {
+    BuckarooView,
+    BuckarooServerMetadata,
+    BuckarooServerMode,
+    pickMode,
+} from "./BuckarooView";
 
-// Defaults are only visible to the widgets if the server sends partial
-// initial_state (it should always send these keys, but useState wants a
-// fully-typed initializer, and the render guard below means the widgets
-// never actually see these values).
-const DEFAULT_DF_META: DFMeta = { total_rows: 0, columns: 0, filtered_rows: 0, rows_shown: 0 };
-const DEFAULT_BUCKAROO_STATE: BuckarooState = {
-    sampled: false,
-    cleaning_method: false,
-    quick_command_args: {},
-    post_processing: false,
-    df_display: "main",
-    show_commands: false,
-};
-const DEFAULT_BUCKAROO_OPTIONS: BuckarooOptions = {
-    sampled: [],
-    cleaning_method: [],
-    post_processing: [],
-    df_display: [],
-    show_commands: [],
-};
-const DEFAULT_COMMAND_CONFIG: CommandConfigT = { argspecs: {}, defaultArgs: {} };
+// Re-export the shared types so existing `import { BuckarooServerMode } from
+// ".../BuckarooServerView"` paths keep working.
+export type { BuckarooServerMetadata, BuckarooServerMode };
 
 /**
  * BuckarooServerView — embed a Buckaroo server session inside a React tree.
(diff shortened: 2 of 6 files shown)


paddymul:
Addressed the codex P2 finding and the CI failure.

**Codex P2 — `fix(js-core): gate BuckarooView render on parquet resolution` (1a2a4c29):**
If a host mounts `BuckarooView` with raw `initial_state` straight off the wire, `df_data_dict` may contain `{ format: "parquet_b64", data: ... }` payload objects. The first render previously seeded state directly from that raw value, so when `df_display_args.main` was also present the widget mounted before resolution ran and `makeStaticInfiniteDs` blew up on `data.slice(...)`.

Fix: detect unresolved parquet payloads up front; when present, initialize `df_data_dict` empty and hold the `Preparing…` guard until `preResolveDFDataDict` completes. Pre-resolved dicts (the `BuckarooServerView` path) skip the effect entirely — no spurious re-render. Added a focused test that asserts the widget stub does **not** appear when the raw dict still has parquet payloads.

**CI failure — `chore: apply paddy-format to autocleaning_pl_test.py` (94f86657):**
The `Python / Lint` job's `paddy-format --check` step has been failing on `main` since #758 was merged (file was committed with formatting that paddy-format wants to fix). It is unrelated to this PR but was blocking it. One-file mechanical reformat — no test changes. Sorry for the carry-over.

PR #765, opened 2026-05-17
feat(xorq-search): plumb search term to JS as highlight_phrase via sd channel

paddymul:
## Summary

Mirrors **#758** for the xorq backend. The xorq path doesn't go through `configure_buckaroo`'s lisp interpreter (ibis exprs can't `.copy()`), so it can't reuse the `SDResult` machinery from #755 directly. Instead this PR adds an analogous sd channel inside `XorqAutocleaning`:

- Handlers in `_XORQ_OP_HANDLERS` may now return either a bare expr (legacy) or `(expr, sd_updates)`. `_apply_xorq_ops` accumulates the per-column sd entries across ops, merging col-by-col.
- `handle_ops_and_clean` runs the accumulated updates through `_rekey_op_sd_to_internal` (the same helper used by `PandasAutocleaning.handle_ops_and_clean` in #758) so the orig-named entries land on buckaroo's internal `a/b/c` letter keys and compose cleanly with the `summary_sd` that `XorqDataflow._get_summary_sd` produces.
- `_xorq_search` returns the filtered expr plus `{col: {'highlight_phrase': [val]}}` for every ibis-String column.

Uses **`highlight_phrase`** (list of literal needles) rather than **`highlight_regex`**: ibis `StringValue.contains` is a literal substring match, so a phrase highlight matches the actual filter semantics.

Scope: only the **search** command is wired today. The sd channel itself is generic — other ops can opt in by returning `(expr, sd_updates)`.

## Test plan

- [x] `pytest tests/unit/test_xorq_buckaroo_widget.py tests/unit/test_xorq_stats_v2.py tests/unit/test_xorq_df_stats_v2.py` — 78 pass locally
- [ ] CI green (xorq jobs)

## Tests added

- `TestSearchHighlight.test_search_op_delivers_highlight_phrase_into_displayer_args` — end-to-end through `XorqBuckarooWidget` with `quick_command_args = {'search': ['admin']}`. Asserts `highlight_phrase == ['admin']` lands in `displayer_args` for both ibis-String columns (`a=name`, `b=role`) and the integer column (`c=score`) stays clean.
- `TestSearchHighlight.test_empty_search_drops_highlight_from_displayer_args` — clearing the search box (`""`) pulls the highlight back out, symmetric to the empty-val short-circuit in `_xorq_search`.

TDD: failing-tests commit was pushed first; CI run on that commit will be visible failing before the implementation commit lands.

### DIFF
--- a/buckaroo/xorq_buckaroo.py
+++ b/buckaroo/xorq_buckaroo.py
@@ -24,7 +24,8 @@ from .buckaroo_widget import BuckarooInfiniteWidget, BuckarooWidget
 from .customizations.styling import DefaultMainStyling, DefaultSummaryStatsStyling
 from .customizations.xorq_stats_v2 import XORQ_STATS_V2
 from .dataflow.autocleaning import (
-    AutocleaningConfig, PandasAutocleaning, generate_quick_ops, merge_ops, ops_eq)
+    AutocleaningConfig, PandasAutocleaning, _rekey_op_sd_to_internal,
+    generate_quick_ops, merge_ops, ops_eq)
 from .dataflow.dataflow import CustomizableDataflow
 from .dataflow.dataflow_extras import Sampling
 from .df_util import old_col_new_col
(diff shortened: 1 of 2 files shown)


paddymul:
Superseded by #774, which rebuilds the same change against main after #767 reorganised the xorq command path. #774's diff is just `buckaroo/customizations/xorq_commands.py` (Search.transform returns `SDResult`) plus the highlight tests — no inline implementation in `xorq_buckaroo.py`, since the SDResult plumbing in the interpreter and `_rekey_op_sd_to_internal` in `PandasAutocleaning` now do that work centrally.

PR #778, opened 2026-05-20
feat(scoped-sd): scoped summary stats (raw + filtered) in merged_sd

paddymul:
## Summary

V1 of the scoped summary stats described in `docs/plans/async-stats.md` (forthcoming): `merged_sd` now carries both the pre-filter baseline (bare keys) and, when a search filter is active, the filtered view (`filtered_*` keys), so the frontend can render both side-by-side via the `?` optional-pinned-row mechanism in #777.

- New trait `summary_sd_raw` holds the bare-key sd. With no filter it aliases `summary_sd` (no extra compute). With a filter active it re-runs autocleaning with empty `quick_command_args` to materialize the unfiltered df, then computes stats on that.
- `_merged_sd` wires `summary_sd_raw` into the bare-key layer and layers `filtered_*` keys from `summary_sd` on top when the filter is on.

## Breaking change

`merged_sd[col]["mean"]` (and the other bare stat names) now refer to the pre-filter dataset. Callers needing the post-filter values should read `filtered_mean` etc. when a filter is active. No in-repo callers needed updating — existing tests construct dataflows without filters, where pre- and post-filter values coincide.

## What's out of V1 (separate follow-ups)

- `cleaned_*` scope (third scope, only meaningful when `cleaning_method != ""`). The current shape proves out raw + filtered first; `cleaned` slots in via a third sd on the cleaned-but-unfiltered df.
- Histogram bin-edge sharing across scopes; categorical↔numerical column-type handling. Tracked in the plan doc.
- Async (`asyncio.to_thread`) compute and two-message WS protocol. V1 is sync.

## Depends on

- #777 (`?` prefix on `PinnedRowConfig`) for the frontend half. Either can land first; combined, the feature lights up.

## Test plan

- [x] 3 new failing-first tests in `tests/unit/dataflow/scoped_summary_stats_test.py` (commit cd2ee5b4) — no-cleaning-no-filter baseline, filter activates `filtered_*` keys, bare keys reflect raw (pre-filter) dataset.
- [x] All 939 Python tests pass locally after the fix commit.
- [ ] CI: verify the failing-tests commit fails and the fix commit passes.

### DIFF
--- a/buckaroo/dataflow/autocleaning.py
+++ b/buckaroo/dataflow/autocleaning.py
@@ -68,7 +68,8 @@ class SentinelAutocleaning:
             cleaned_df = SENTINEL_DF_2
         else:
             cleaned_df = df
-        return [cleaned_df, {}, generated_code, merged_operations]
+        # Sentinel stub: filtered and unfiltered views are identical.
+        return [cleaned_df, {}, generated_code, merged_operations, cleaned_df]
 
 class AutocleaningConfig:
     command_klasses = [DefaultCommandKlsList]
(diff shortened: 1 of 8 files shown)


paddymul:
Superseded by #785, which uses #783's keyed SD cache as the substrate instead of the parallel 5-tuple handle_ops_and_clean shape this PR took. Closing.

PR #787, opened 2026-05-20
spike(rows-first): two-message state_change protocol behind env gate

paddymul:
**Draft / spike** — not for merge. Validates the wire shape for *"show rows first, compute stats after"* without going async. Synchronous server, just message reordering + an ``IOLoop.call_later`` yield.

Sits on top of #785 (uses ``_defer_summary_sd`` + ``recompute_summary_sd`` hooks that this PR adds to the dataflow).

## What it proves

- A single ``buckaroo_state_change`` produces **two** ``initial_state`` frames on the wire.
- An ``infinite_request`` fired by the client between the two frames is serviced and its parquet rows arrive **before** phase 2. The ``call_later(0.01)`` between phase 1 and phase 2 gives inbound socket reads a real time window to win the race.

Test: ``tests/unit/server/test_rows_first_spike.py`` (2 cases).

## How to try it

Set the env flag and exercise normally:

```bash
BUCKAROO_ROWS_FIRST_SPIKE=1 uv run ...
```

Default (flag off) keeps today's single-frame rebroadcast. Existing ``test_load_expr`` tests pass unchanged.

## Open questions

- **xorq path coverage.** ``XorqServerDataflow`` may not route through the ``_defer_summary_sd`` short-circuit on the base ``DataFlow``. First spike test had to drop a "stats actually differ between phases" assertion because it was failing on xorq. **This is the most important item to resolve before un-gating** — without that assertion the spike test would still pass if ``recompute_summary_sd()`` were a no-op, so we can't detect regression of the spike's whole point.
- **Inter-phase ``state_change`` race.** Between phase 1 broadcast and the ``call_later`` firing, a second ``state_change`` can land. Phase 2 then computes stats against a ``processed_df`` that already reflects the *second* state, but the broadcast payload mixes phase-2 stats with what the user thinks is the first interaction. The spike doesn't try to address this; needs a serialise / cancel-previous strategy before un-gating.
- **Delay tuning.** 10ms is a localhost-only number — loopback WS RTT is well under 1ms. For remote deployments (RTT 30-100ms+) 10ms is *shorter* than the round-trip and the spike's premise breaks. Real prod tuning must measure end-to-end and pick a value that reliably loses the race to a network-bound ``infinite_request``.
- **Cache-hit fast path.** When ``filt_sd_key`` for the new state is already in ``summary_stats_cache`` (search-box backspace, oscillation), there's no compute to defer — phase 2 would be near-instant. The spike doesn't short-circuit and emits two frames anyway, which is wasteful.
- **Phase-2 message type.** Both phases currently send ``type: "initial_state"``. JS will re-apply full state twice. Worth deciding whether phase 2 should be a distinct type (e.g. ``stats_update``) so the front-end can target a cheaper re-render of just the stats-driven UI, rather than re-running the full ``initial_state`` handler.
- **No before/after timing.** PR currently asserts the *protocol* works but doesn't measure whether it produces a perceptible win on a realistic dataset. One A/B measurement (rows-rendered latency, with/without flag, on a search that triggers a real compute) would justify or kill the protocol.

## Stack

- #777 ✅ merged (JS ``?key`` pinned-rows)
- #783 ✅ merged (keyed SD cache)
- #785 ⏳ open (merge raw + filt scope SDs into prefixed ``merged_sd``)
- **This** ⏳ draft (rows-first two-message protocol; depends on #785)

### DIFF
--- a/buckaroo/dataflow/dataflow.py
+++ b/buckaroo/dataflow/dataflow.py
@@ -250,6 +250,27 @@ class DataFlow(ABCDataflow):
 
     _summary_sd_cache_key = (None, None)
 
+    # Spike (rows-first WS protocol): when set, the next firing of the
+    # ``_summary_sd`` observer is short-circuited so the cascade lands
+    # ``processed_df`` + ``df_meta`` (the cheap parts) without paying the
+    # analysis-pipeline cost. Caller is responsible for clearing the flag
+    # and re-running ``_summary_sd`` via an explicit ``recompute_summary_sd()``
+    # call once the row-data has been shipped. **Not for production paths.**
+    _defer_summary_sd = False
+
+    def recompute_summary_sd(self):
+        """Force a (re-)compute of ``summary_sd`` against the current
+        ``processed_df``. Used by the spike state-change handler to lift
+        the ``_defer_summary_sd`` short-circuit after the row-data has
+        been pushed to the client.
+        """
+        df = self.processed_df
+        if df is None:
+            return
+        result_summary_sd, errs = self._get_summary_sd(df)
+        self.summary_sd = result_summary_sd
+        self.errs = errs
+
     @observe('processed_result', 'analysis_klasses')
     @exception_protect('summary_sd-protector')
     def _summary_sd(self, change):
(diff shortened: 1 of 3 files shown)


paddymul:
Closing per triage: superseded by the JS-driven progressive-stats design (#808/#809/#810). plans/0787-xorq-rows-first-coverage.md (in #790) commits to that path.

PR #855, opened 2026-05-24
feat(server): /load accepts backend='polars' for buckaroo dataflow

paddymul:
## Summary
- `/load` was hardcoded to pandas (`load_file` → `create_dataflow`); the polars buckaroo dataflow was only reachable via the notebook widget. Demos that embed via the standalone server (iframe / `BuckarooServerView`) couldn't exercise it.
- Adds `backend: 'polars'` to the POST body (default `'pandas'`, polars validated as `mode='buckaroo'`-only). New module `data_loading_polars.py` with `PolarsServerDataflow` (config lifted from `PolarsBuckarooInfiniteWidget`), eager polars file loader, and `handle_infinite_request_buckaroo_polars` (literal-substring `str.contains` across String columns to match pandas `search_df_str` semantics).
- WS dispatch routes `session.backend == 'polars'` to the new handler; pandas / xorq branches untouched. Polars stays optional — `data_loading_polars` is lazy-imported.

Usage:
```python
POST /load { "session": "demo", "path": "data.parquet", "mode": "buckaroo", "backend": "polars" }
```

## Test plan
- [ ] `pytest tests/unit/server/` — existing pandas-backend `/load` tests still pass (92 passed locally).
- [ ] Manual: POST /load with backend=polars, then open the session — grid renders, sorting and search work end-to-end.

### DIFF
--- a/buckaroo/server/data_loading_polars.py
+++ b/buckaroo/server/data_loading_polars.py
@@ -1,0 +1,135 @@
+"""Polars counterparts of the pandas-based loaders in ``data_loading``.
+
+Lives in its own module so polars stays an optional dependency — the
+server only imports this when ``/load`` is called with
+``backend: "polars"``. Mirrors the shape of ``data_loading``:
+
+* :func:`load_file_polars` reads parquet/csv/tsv/json eagerly into a
+  ``pl.DataFrame``.
+* :class:`PolarsServerDataflow` is the polars analogue of
+  ``ServerDataflow`` — same role in the pipeline, polars-flavored
+  analysis / autocleaning / stats / sampling classes (lifted from
+  ``PolarsBuckarooInfiniteWidget``).
+* :func:`handle_infinite_request_buckaroo_polars` is the polars
+  equivalent of ``handle_infinite_request_buckaroo`` — applies the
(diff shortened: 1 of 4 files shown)


paddymul:
@codex thanks — both P1 and P2 were addressed in [34deb3c0](https://github.com/buckaroo-data/buckaroo/commit/34deb3c0):

- **P1 (no string columns)** — `handle_infinite_request_buckaroo_polars` now uses `processed_df.clear()` to return an empty frame when `search_string` is non-empty and the frame has no `String` columns. Matches pandas `search_df_str` semantics (OR-accumulate from an all-False mask). Pinned by `test_search_with_no_string_columns_returns_empty`.
- **P2 (.json vs ndjson)** — `.json` now dispatches to `pl.read_json` (standard JSON arrays, matching `pd.read_json` default `lines=False`); newline-delimited JSON moved to the explicit `.ndjson` extension. Pinned by `test_load_file_polars_json_array` + `test_load_file_polars_ndjson_still_works`.

Lint failure in the prior CI run was unrelated — `paddy_format` rejected the multi-line import block in the new test file. Fixed in [8eded229](https://github.com/buckaroo-data/buckaroo/commit/8eded229).

PR #863, opened 2026-05-28
feat(server): /load_expr accepts column_config_overrides, extra_grid_config, init_sd

paddymul:
## Summary

Adds styling and grid config customization support to `/load_expr` endpoint, matching the functionality available on `/load` (PR #850):

- Extract `column_config_overrides`, `extra_grid_config`, `init_sd` from request body
- Pass these to `XorqServerDataflow` constructor
- Reset each WebSocket client's `search_string` on dataset change (same fix as #851)

This allows xorq-backed dataflows to be styled and configured identically to pandas/polars-backed dataflows via the same parameters.

## Test plan

- [ ] Full build passes: `./scripts/full_build.sh`
- [ ] Python tests pass: `pytest -vv tests/unit/`
- [ ] Verify xorq backend accepts and applies config overrides (test with xorq build dir + styling config)
- [ ] Verify search_string resets on /load_expr call (old filters don't carry over)

### DIFF
--- a/buckaroo/server/handlers.py
+++ b/buckaroo/server/handlers.py
@@ -419,6 +419,9 @@ class LoadExprHandler(tornado.web.RequestHandler):
         prompt = body.get("prompt", "")
         no_browser = bool(body.get("no_browser", False))
         component_config = body.get("component_config")
+        column_config_overrides = body.get("column_config_overrides")
+        extra_grid_config = body.get("extra_grid_config")
+        init_sd = body.get("init_sd")
 
         project_root = body.get("project_root")
         cache_storage_path = body.get("cache_storage_path")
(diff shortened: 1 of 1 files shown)


paddymul:
Rebased onto main (picked up #854 per-client search_string and #855 polars backend). The second original commit (`6dbaca83`) — which removed `client.search_string = ""` from LoadExprHandler — was dropped during rebase: that removal was correct against the old session-level search_string but wrong against main's per-client model, so main's per-client reset is the right behaviour to keep.

Code review (medium effort, all 7 angles): **no findings**. The three new kwargs (`column_config_overrides`, `extra_grid_config`, `init_sd`) flow correctly through `XorqServerDataflow(**kwargs)` → `CustomizableDataflow.__init__` (which declares them as strict named params — typos raise `TypeError`). All three reach `df_display_args` via `CustomizableDataflow`'s existing machinery. Pre-existing patterns (broad `except` in LoadExprHandler, inline push loop vs `_push_state_to_clients`) were not introduced by this diff.

PR #871, opened 2026-05-29
ci: merge npm release into PyPI release workflow

paddymul:
## Summary

- Adds a `release-npm` job to `release.yml` that runs automatically after the `release` (PyPI) job completes
- Version flows from the PyPI job via `jobs.release.outputs.version` — no manual coordination needed
- The npm job inherits the same idempotency check (skips if that version is already on npm)
- `release-npm.yml` is kept for emergency standalone re-publishes

## One manual step required

The npm Trusted Publisher on npmjs.com must be updated:

- Go to https://www.npmjs.com/package/buckaroo-js-core/access
- Change the registered workflow from `release-npm.yml` → `release.yml`
- Environment stays `npm`

The OIDC diagnostics step prints the exact values npm checks, so if auth fails it's easy to compare against the registered config.

### DIFF
--- a/.github/workflows/release-npm.yml
+++ b/.github/workflows/release-npm.yml
@@ -1,11 +1,11 @@
 name: Release npm
 
-# Standalone workflow that publishes packages/buckaroo-js-core to npm.
-# Kept separate from Release (PyPI) so we can iterate on it without
-# touching the wheel/PyPI pipeline. To keep buckaroo-js-core in lockstep
-# with the Python package, run this with `version` set to the same
-# semver as the matching Python release (e.g. the git tag created by
-# the Release workflow).
+# Emergency standalone workflow for re-publishing buckaroo-js-core to npm.
+# The normal path is the release-npm job in release.yml, which runs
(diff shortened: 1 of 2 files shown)


paddymul:
## Review follow-up (043feb9e)

Three findings from the code review, all addressed:

**`if: needs.release.result == 'success'` guard** — added to the `release-npm` job. Without this, a skipped or cancelled PyPI job would still trigger the npm job, which would then call `npm version ""` with an empty string and fail with a confusing error rather than a clean skip.

**Non-deterministic `npm@latest` pin** — pinned to `npm@11.5.1` (the minimum version that satisfies Trusted Publishing) in both `release.yml` and `release-npm.yml`. `npm@latest` could pick up a breaking release silently on any future run.

**Stale `release-npm.yml` comment block** — updated the header to describe the file's actual current role: emergency standalone re-publisher, not the primary publish path. The old comment ("kept separate so we can iterate without touching PyPI") no longer reflected reality after the merge.

PR #882, opened 2026-05-31
feat(server): /load backend=polars, /load_expr config parity, per-client search isolation (#851)

paddymul:
## What this does

Bundles a set of standalone-server `/load` enhancements that were sitting uncommitted on `feat/load-column-config-overrides`. They extend the server so a `POST /load` (and `/load_expr`) call can reach the same configuration surface a notebook `BuckarooInfiniteWidget` already has, add a polars loading path, and fix a multi-client live-search bug.

### 1. `backend='polars'` for `POST /load`
New module `buckaroo/server/data_loading_polars.py`:
- `load_file_polars` / `get_metadata_polars` — eager parquet/csv/tsv/json read into a `pl.DataFrame`.
- `PolarsServerDataflow` — polars analogue of `ServerDataflow` (polars analysis / autocleaning / stats / sampling), capped at `pre_limit=1_000_000` rows for the stats pipeline.
- `handle_infinite_request_buckaroo_polars` — polars row-fetch handler; applies the live `search_string` as a literal substring match on String columns, matching the pandas `search_df_str` semantics.

`handlers.py` validates `backend` (`pandas` default, or `polars`; `polars` only valid with `mode='buckaroo'`) and routes the load/build path. `websocket_handler.py` routes the infinite row-fetch to the polars handler when `session.backend == 'polars'`. **polars stays an optional dependency** — it's imported lazily, only when a request actually asks for `backend='polars'`.

### 2. Per-client `search_string` (fixes #851)
`search_string` was stored on the shared `SessionState`, so two browser clients on the same session clobbered each other's live-typed filter (and each other's highlight). It now lives on the `DataStreamHandler` instance:
- The row-fetch dispatch reads `self.search_string` instead of `session.search_string`.
- A new `_send_highlight_overlay` sends an `initial_state` with `highlight_phrase` injected into string-column `displayer_args` to **only the typing client** — never broadcast — and is skipped when a dataflow rebuild is already going to broadcast the highlight anyway.
- The term is stripped before snapshotting `buckaroo_state` onto the session, and reset to `""` on every new `/load` / `/load_expr` push.

### 3. `/load_expr` config-override parity
`LoadExprHandler` now forwards `column_config_overrides`, `extra_grid_config`, and `init_sd` to `XorqServerDataflow`, mirroring the `/load` kwargs added in `6bf4c12b`.

### 4. `init_sd` vs. highlight ordering (`styling.py`)
`init_sd` `displayer_args` are now merged **before** lowcode-op highlight metadata is injected, so a column whose pandas type is `obj` but which `init_sd` promotes to `displayer: 'string'` still picks up `highlight_phrase`. The highlight injection skips any key the caller already set, so an explicit `init_sd` highlight wins.

### Tangential (also uncommitted, rode along)
- `pyproject.toml` / `uv.lock`: add `matplotlib` (used by `buckaroo/customizations/histogram.py`).
- `packages/buckaroo-js-core/src/stories/DFViewerDirect.stories.tsx`: a Storybook story demonstrating the direct `<DFViewer>` consumer pattern.

## Base note
This PR is based on `main` but **includes commit `6bf4c12b`** (`/load accepts column_config_overrides, extra_grid_config, init_sd`) in addition to the new work, because the polars/load_expr changes build directly on that plumbing and that commit is not yet in `main`. If `6bf4c12b` lands separately first, rebasing will collapse this PR's diff to just the new commit.

## Testing
- `pytest tests/unit/server/` — 103 passed, 1 skipped (the one failure, `test_mcp_uvx_install`, is a network `uvx` install that timed out — unrelated).
- `pytest tests/unit/.../customizable_dataflow_test.py tests/unit/server/test_server.py -k "highlight or init_sd or string"` — 6 passed.
- `ruff check` + `paddy-format --check` (full tree) — clean.
- No dedicated tests for the polars backend or the #851 per-client behaviour yet — worth a follow-up.


paddymul:
Fixed two issues in ce3955bc:
- **matplotlib removed**: it's not imported anywhere in the package (histogram.py only referenced it in a comment URL). Removed from core dependencies and uv.lock entirely.
- **JSON reader fixed**: changed `pl.read_ndjson` to `pl.read_json` for the `.json` extension — standard JSON arrays (`[{...}]`) need `read_json`, not the NDJSON reader.

paddymul:
Closing as superseded. All features from this PR landed in main via smaller focused PRs: per-client search isolation (#854), polars dataflow (#855, #861), /load_expr config parity (#863), init_sd/highlight ordering (#853), and the Storybook story (#856). Main has also moved ahead with fixes this branch is missing (search_string round-trip in build_state_message, build_dir/project_root on SessionState, skip_stat_columns/cache_storage_path on LoadExprHandler).

PR #890, opened 2026-06-05
feat(js): add "fitContent" layout mode (self-sizing widget box)

paddymul:
## Problem

A fixed-height host that embeds Buckaroo (e.g. a `60vh` wrapper) gets a dark gap below short tables. Buckaroo sizes its grid to `window.innerHeight / 2` and fills its container with `height: 100%`, so when the content is shorter than the container, the leftover space shows through. The [height-modes docs](https://buckaroo-data.readthedocs.io/en/latest/articles/height-modes.html) tell hosts to make "the outer container match the inner viewport" — that manual coordination is the thing that breaks when a host picks its own size.

## Change

Adds a third `component_config.layoutType` value, **`"fitContent"`**, alongside `"autoHeight"` and `"normal"`:

- **Grid:** behaves exactly like auto-detect — `autoHeight` when the rows fit, `normal` + virtualized scroll (capped at `dfvHeight`) when they don't. `heightStyle` maps `fitContent` to that and never emits it as a literal AG Grid `domLayout`.
- **Outer box:** the widget wrappers drop their `height: 100%` when the layout is `fitContent`, so the widget sizes its own box to its content. The host gives it an `auto` / `max-height` wrapper and gets no gap — without having to match the viewport, and without reaching into AG Grid's DOM.

Wired into **both** modes: `BuckarooInfiniteWidget` (`"buckaroo"`) and `DFViewerInfiniteDS` (`"viewer"`), plus `BuckarooView`'s `anywidget` wrapper. New `isFitContentLayout()` helper.

This replaces the earlier `onHeightChange` callback approach (an earlier version of this PR): a self-sizing layout needs no height reported back to the host, and — unlike a JS callback — it works the same in Jupyter/anywidget, where JS can't call back into Python.

## Tests

- **jest** (`gridUtils.test.ts`, runs in `JS / Build + Test`): `fitContent` routes to `autoHeight`/`normal`, never emits `"fitContent"` as a domLayout, doesn't throw without an explicit `dfvHeight`.
- **Playwright/Storybook** (`fitcontent-height.spec.ts` + `FitContentHeight` story, registered in `scripts/test_playwright_storybook.sh`): the widget (5 rows) in a 600px container collapses well below 600px and ends flush with its content (no gap).

Per TDD, the first commit lands both test layers failing on CI; the implementation follows.

### DIFF
--- a/packages/buckaroo-js-core/pw-tests/fitcontent-height.spec.ts
+++ b/packages/buckaroo-js-core/pw-tests/fitcontent-height.spec.ts
@@ -1,0 +1,48 @@
+/**
+ * Playwright test for the `layoutType: "fitContent"` height mode.
+ *
+ * The FitContentHeight story mounts the widget (5 rows) in a 600px container.
+ * In fitContent mode the widget sizes its own box to its content, so:
+ *   1. The widget root (.buckaroo-widget) is far shorter than the 600px host.
+ *   2. There is no gap between the bottom of the table and the bottom of the
+ *      widget (the widget ends flush with its content).
+ */
+import { test, expect } from "@playwright/test";
+import { waitForCells } from "./ag-pw-utils";
+
+const STORY_URL =
+  "http://localhost:6006/iframe.html?viewMode=story&id=buckaroo-fitcontentheight--primary&globals=&args=";
+
+const CONTAINER_HEIGHT = 600;
+
+test.describe("fitContent layout mode", () => {
+  test("widget sizes to its content, not the 600px host container", async ({ page }) => {
+    await page.goto(STORY_URL);
+    await waitForCells(page);
+
+    const dataGrid = page.locator(".df-viewer");
+    await expect
+      .poll(async () => (await dataGrid.textContent())?.includes("Alice"), { timeout: 10_000 })
+      .toBe(true);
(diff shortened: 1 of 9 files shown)


paddymul:
## Review summary

**Code review:** Clean and correct. The `fitContent` layout mode is well-implemented:

- `isFitContentLayout` correctly centralises the check; the outer wrappers in both `BuckarooInfiniteWidget` and `DFViewerInfiniteDS` apply it symmetrically.
- `gridUtils.ts` maps `fitContent → undefined` before passing to AG Grid so `"fitContent"` never leaks as a `domLayout` value.
- `BuckarooView.tsx`'s `contentSized` flag correctly treats `fitContent` the same as `autoHeight` for the outer wrapper.
- Tests cover all four routing cases (fits/doesn't fit × with/without `dfvHeight`).

**Codex P1 comment on `useReportContentHeight.ts`** (discussion_r3364611226): stale. The branch was force-pushed to replace the `onHeightChange` callback approach entirely. `useReportContentHeight.ts` and the `WidgetHeightReport` story no longer exist in the current branch — the `fitContent` layout mode is the replacement, and it needs no JS callback. No action required.

PR #892, opened 2026-06-06
feat(search): make search debounce configurable via component_config (#891)

### LINKED ISSUE #891: Make the search debounce configurable across BuckarooServerView / BuckarooView / BuckarooInfiniteWidget / BuckarooWidget
## Summary

The quick-search debounce is hardcoded with no prop/option to change it per embedding. The search box
dispatches the term as ``quick_command_args.search``, which the server treats as a dataflow mutation
(``_DATAFLOW_FIELDS``, ``server/websocket_handler.py:28``), so every completed search rebuilds the full
summary-stat pipeline before any rows come back. On expensive backends that rebuild dominates, and an
embedder has no way to lengthen the debounce so a burst of keystrokes coalesces into one search.

Measured against a live diff session (an outer-join compare expression over xorq — 500 rows, ~16
columns including the hidden ``membership`` / ``*_eq`` / ``*_pct_delta`` machinery):

- ``buckaroo_state_change`` (the stat-pipeline rebuild): **~0.55 s** for a zero-match term, **~1.7 s**
  for a 100+-match term (``"West"``, 104/500 rows). Repeated 5× for the same term with no warm-up —
  1754, 1884, 1688, 1697, 1766 ms — the filtered stats are recomputed from scratch each time.
- The follow-on ``infinite_request`` that actually filters and returns rows: **~65 ms**.

So the row filter is cheap; the per-search stat rebuild is the cost, it scales with matched-row count
(not table size), and it is not cached. Per keystroke that is unusable on a slow backend.

The only debounce knobs that exist today (``StateOrchestrator`` ``minDebounceMs`` / ``maxDebounceMs`` /
``multiplier``, ``server/StateOrchestrator.ts``) are internal to the progressive-stats scheduler and
are not surfaced on any React component — ``BuckarooServerViewProps`` and ``BuckarooViewProps`` expose
no debounce field.

## Suggested fix

Add a ``searchDebounceMs`` option (number, ms) threaded through every embedding entry point, defaulting
to the current behavior:

- ``BuckarooInfiniteWidget`` — where the search input lives; this is the actual debounce site.
- ``BuckarooView`` → ``BuckarooServerView`` — pass-through props so WS/IPC hosts can set it.
- ``BuckarooWidget`` (Python anywidget) — a traitlet plumbed into ``component_config`` so Jupyter
  embeds and server-rendered sessions can set it too.
- ``WidgetDCFCell`` — same pass-through.

Where practical, unify this with the ``StateOrchestrator`` min/max debounce parametrization that already
exists rather than introducing a second independent timer.

## Impact

An embedder can set, e.g., ``searchDebounceMs={2000}`` on a diff/compare view so the expensive stat
rebuild fires once after the user stops typing instead of once per keystroke, while every other embed
keeps the snappy default. No server change required.

## Scope

``components/BuckarooWidgetInfinite.tsx`` (search input + debounce), ``server/BuckarooView.tsx``,
``server/BuckarooServerView.tsx``, ``server/StateOrchestrator.ts``, and the Python side
(``buckaroo_widget.py`` traitlet → ``component_config``).

## Context

Identified while diagnosing slow search on a pydata-app diff view. Numbers above are from a real
``/load_expr`` compare session over the cbl-trip-distance dataset.

Related: the server already has a per-client live ``search_string`` filter path that skips the stat
pipeline entirely (``handle_infinite_request_xorq``, ``server/xorq_loading.py``; #838), but no released
buckaroo-js-core emits ``search_string`` — so today every search goes through the heavy
``quick_command_args.search`` path. A configurable debounce is the smaller, immediate mitigation;
wiring the frontend to the ``search_string`` live path would be the larger fix.

paddymul:
## Summary

- Adds `searchDebounceMs` to Python's `ComponentConfig` TypedDict in `styling_core.py`
- `StatusBar` accepts a new `componentConfig: Record<string, unknown>` prop and places it in its AG-Grid `context`
- `fakeSearchCell` reads `params.context.componentConfig.searchDebounceMs` for the keystroke debounce, falling back to the existing 300 ms default

## Usage

```python
BuckarooInfiniteWidget(df, component_config={'searchDebounceMs': 2000})
```

Any field added to Python's `ComponentConfig` is now automatically available to AG-Grid cell renderers in the status bar via `params.context.componentConfig` — no prop threading required.

## Test plan

- [ ] JS build passes (`pnpm run build`)
- [ ] 284 jest tests pass (`pnpm test`)
- [ ] Manually verify: set `searchDebounceMs=2000`, type in search box, confirm dispatch waits 2 s after last keystroke
- [ ] Confirm default behavior (300 ms) is unchanged when `searchDebounceMs` is not set

Closes #891

### DIFF
--- a/buckaroo/dataflow/styling_core.py
+++ b/buckaroo/dataflow/styling_core.py
@@ -167,7 +167,8 @@ ComponentConfig = TypedDict('ComponentConfig', {
     'shortMode': NotRequired[bool],
     'selectionBackground': NotRequired[str],
     'className': NotRequired[str],
-    'theme': NotRequired[ThemeConfig]})
+    'theme': NotRequired[ThemeConfig],
+    'searchDebounceMs': NotRequired[int]})
 
 DFViewerConfig = TypedDict('DFViewerConfig', {
     'pinned_rows': List[PinnedRowConfig],
--- a/packages/buckaroo-js-core/src/components/BuckarooWidgetInfinite.tsx
+++ b/packages/buckaroo-js-core/src/components/BuckarooWidgetInfinite.tsx
@@ -338,7 +338,7 @@ export function BuckarooInfiniteWidget({
                         buckarooOptions={buckaroo_options}
                         themeConfig={cDisp.df_viewer_config?.component_config?.theme}
                         inFlight={inFlight}
-                        componentConfig={cDisp.df_viewer_config?.component_config as Record<string, unknown> | undefined}
+                        componentConfig={effectiveDisplayArgs['main']?.df_viewer_config?.component_config as Record<string, unknown> | undefined}
                     />
                     <DFViewerInfinite
                         key={effectiveDataframeId}
--- a/packages/buckaroo-js-core/src/components/BuckarooWidgetInfinite.tsx
+++ b/packages/buckaroo-js-core/src/components/BuckarooWidgetInfinite.tsx
@@ -338,6 +338,7 @@ export function BuckarooInfiniteWidget({
                         buckarooOptions={buckaroo_options}
                         themeConfig={cDisp.df_viewer_config?.component_config?.theme}
                         inFlight={inFlight}
+                        componentConfig={cDisp.df_viewer_config?.component_config as Record<string, unknown> | undefined}
                     />
                     <DFViewerInfinite
                         key={effectiveDataframeId}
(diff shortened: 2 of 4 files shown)


paddymul:
## Code review summary

**Codex finding:** `componentConfig` was read from `cDisp` (the currently selected display). Since `component_config` is only installed on the `main` display entry, switching to summary or any other display silently passed `undefined` to `StatusBar`, dropping the configured `searchDebounceMs` back to 300 ms.

**Fix applied (7f3be7cd):**
- `BuckarooWidgetInfinite.tsx`: changed to `effectiveDisplayArgs['main']?.df_viewer_config?.component_config`
- `DCFCell.tsx`: changed to `df_display_args['main']?.df_viewer_config?.component_config`

Both the build and 284 Jest tests pass with the fix.

Issue #843, opened 2026-05-24
perf: DefaultSummaryStats O(n²) on object columns with unhashable values (lists/dicts/sets)


paddymul:
## Summary

`DefaultSummaryStats.series_summary` calls `ser.value_counts()` and `get_mode(ser)` unconditionally. For object-dtype Series whose values are unhashable (Python `list`, `dict`, `set`), pandas can't hash the values and falls back to an O(n²) pairwise comparison. A 5000-row `list<string>` column takes ~600 ms in `value_counts` alone, and per-handler profiling attributes 99% of `create_dataflow` cost to this one call.

## Repro on `main`

```python
import time, random, warnings; warnings.filterwarnings('ignore')
import pandas as pd, numpy as np
random.seed(42); np.random.seed(42)
from buckaroo.server.data_loading import create_dataflow

def make_nested_lists(n):
    words = ['alpha','beta','gamma','delta','epsilon']
    return pd.DataFrame({'tags': [random.choices(words, k=30) for _ in range(n)]})

for n in (1000, 2500, 5000, 10000):
    df = make_nested_lists(n)
    t0 = time.perf_counter(); create_dataflow(df); dt = time.perf_counter() - t0
    print(f'n={n:>5}  {dt*1000:>7.0f} ms')
```

```
n= 1000       30 ms
n= 2500      145 ms     (~5x for 2.5x rows)
n= 5000      602 ms     (~4x for 2x rows)
n=10000     2545 ms     (~4x for 2x rows)
```

Isolating `value_counts` on the same list column confirms the same scaling:

```
n= 1000  value_counts=   20 ms
n= 2500  value_counts=  144 ms
n= 5000  value_counts=  651 ms
n=10000  value_counts= 2743 ms
```

## Per-handler attribution (n=2000)

```
DefaultSummaryStats__series           90.5 ms   ← 99%
ComputedDefaultSummaryStats__computed  0.3 ms
Histogram__computed                    0.2 ms
TypingStats / others                  ~0 ms
```

## Root cause

`buckaroo/customizations/analysis.py:131,139`:

```python
@staticmethod
def series_summary(sampled_ser, ser):
    l = len(ser)
    value_counts = ser.value_counts()      # ← O(n²) on unhashable values
    ...
    base_d = dict(
        ...
        value_counts=value_counts,
        mode=get_mode(ser),                # ← also O(n²)
        ...)
```

[...]

PR #912, opened 2026-06-10
Remove all vestiges of dfstats_v1 (full @stat port)

paddymul:
Removes the entire legacy v1 DfStats / `ColAnalysis`-execution stack now that the v2 `@stat` `StatPipeline` is the only analysis path. (Replaces #876 — rebased onto current `main`.)

## Removed
- **v1 executor**: `DfStats` / `AnalysisPipeline` (deleted `analysis_management.py`), `PlDfStats` / `PolarsAnalysisPipeline` + the polars `produce_*` helpers, and the v1 `order_analysis` / `check_solvable` ordering.
- **v1 adapter**: `v1_adapter.py`, the `ColAnalysis` branch in `_normalize_inputs`, and the `v1_computed` / `spread_dict_result` `StatFunc` flags.
- **compat shims**: `process_df_v1_compat` / `process_table_v1_compat` / `_find_v1_class`. The `DfStatsV2` / `PlDfStatsV2` / `XorqDfStatsV2` wrappers now call `process_df` / `process_table` and build the `ErrDict` via a new `errors_to_errdict`.
- **dead v1 `ColAnalysis` stat classes**: `TypingStats`, `DefaultSummaryStats`, `ComputedDefaultSummaryStats`, `PdCleaningStats`, `Histogram`, `HeuristicFracs`, the cleaning-op-gen classes, and `customizations/heuristics.py` (the shared helper functions are kept).

## Ported to @stat
- `BuckarooWidget`, the headless server, and the pandas autocleaning configs run on `PD_ANALYSIS_V2`; added a `cleaning_gen_ops` `@stat` for the default autoclean op.
- `_normalize_inputs` accepts structural `ColAnalysis` classes (styling, post-processing) as no-ops and rejects leftover v1 stat classes with a clear error.
- All affected tests ported to `@stat` / `StatPipeline.process_df`; v1-mechanic and v1-vs-v2 parity tests dropped where covered by `test_paf_v2` / `test_pd_stats_v2`.

## Polars autocleaning (on v2)
Re-implemented on the v2 pipeline (it previously only worked through the v1 `PlDfStats` executor): `pl_cleaning_stats` `@stat` + `PL_AUTOCLEAN_DEFAULT_V2`, a re-added `PolarsAutocleaning` (`PlDfStatsV2` + polars `make_origs`), and `PolarsBuckarooWidget` wired to it.

## Two latent regressions fixed (exposed by routing the pandas widget through @stat)
- `base_summary_stats` short-circuits `value_counts` on unhashable (list/dict/set) columns — restores #843.
- `make_origs` checks `add_orig` by truthiness, so `@stat` cleaning funcs returning `add_orig=False` no longer add spurious `_orig` columns.

## Notes
- The `docs/example-notebooks/mo-autocleaning.py` sample still references a removed v1 cleaning class; it's a docs notebook (not imported, not in CI).
- Full unit suite green (911 passed, 0 failed); `ruff` + `paddy_format --check` clean.


paddymul:
## Code review summary

Reviewed the full v1-removal diff (47 files, +840/−3278). The removal is faithful and well-guarded; I found one correctness regression that this PR ships, fixed it under the test-first flow, and noted a few smaller items. CI is now green on `a0e2c5db`, including the `Python / Test (3.14)` job.

### Fixed

**Lost `np.histogram` ValueError guard (the one real regression).** The v1 `Histogram.series_summary` wrapped `np.histogram(meat, 10)` in `except ValueError` for integers near 2^53, where float64 can't produce 10 distinct bin edges (added in f2147ce9 alongside `bigint-test.html`). The v2 `histogram_series` / `pl_histogram_series` dropped the guard. Because this PR makes the v2 `@stat` path the default pandas widget path, routing a bigint column through `DfStatsV2` now raised `ValueError`, printed reproduction noise on every display, and left `histogram: None` (no histogram rendered). Reproduced end-to-end with the repo's own bigint data (`scripts/generate_static_test_html.py:28`). CI stayed green before because the bigint Playwright test asserts grid cells, not that column's histogram.

- `3ec49f69` — failing tests: direct-call contract for `histogram_series` / `pl_histogram_series` plus the `PD_ANALYSIS_V2` pipeline end-to-end (seen failing on run 27370896726).
- `55176fba` — restored the `except ValueError` guard in both functions, returning empty `histogram_args` / `histogram_bins` so the categorical fallback renders.
- `804a713a` — made the assertions numpy-version-agnostic (numpy 2.0.2 returns near-degenerate bins instead of raising, so the test asserts the contract — no raise, correct result shape — rather than an empty dict).

### CI flake fixed along the way (unrelated to this PR)

`tests/unit/perf_smoke_test.py::test_slow_stat_flagged` (added to main in #917) failed only on Python 3.14 via the PR merge ref. Its 2000-row O(n²) stat ran just under the 0.25s `time_floor_seconds` on 3.14's faster interpreter, so nothing got flagged. It runs its own quadratic stat and never touches this PR's code.

- `e82dd617` — merged current `main` into the branch (it was behind; clean auto-merge).
- `a0e2c5db` — bumped the test to 4000 rows so the per-column call clears the floor across interpreter speeds (measured ~16x margin locally → ~4x on the 3.14 runner). Default floor unchanged; the recorder short-circuits after the first column, so the test still finishes in a few seconds.

Two judgment calls for the author: this perf-test change is unrelated to the PR and could move to its own commit against main (the 3.14 marginality exists on main too); and `main` was merged in rather than rebased. Happy to split or rebase if preferred.

### Noted, not changed

- **Stale comment in `autocleaning.py` (~line 255).** It says the polars `make_origs` "walks cleaning_sd keys as actual column names," justifying the post-`make_origs` rekey. This PR changed polars `make_origs` to index by the `orig_col_name` field, so that ordering is no longer load-bearing. Worth updating so the constraint doesn't read as still-required.
- **pandas/polars autoclean asymmetry.** `PD_AUTOCLEAN_DEFAULT_V2` includes `typing_stats` / `_type` / `numeric_stats`; `PL_AUTOCLEAN_DEFAULT_V2` is just `[pl_cleaning_stats, cleaning_gen_ops]`. Worth a deliberate call on whether polars needs `_type` for cleaning-detail styling, or pandas is computing extra stats per cleaning pass.
- **Duplicate `StatError` per multi-key stat (pre-existing).** A multi-key stat that fails records one `StatError` per provided key; `errors_to_errdict` collapses them but `print_errors` paths show duplicates. Not introduced here.

### Done well

`_normalize_inputs` is the riskiest change and it's careful: structural styling classes pass as no-ops, while anything still carrying v1 stat surface (`series_summary`/`computed_summary` overrides, `select_clauses`, `column_ops`, non-empty `provides_defaults`) raises a clear `TypeError` instead of silently dropping stats — all four rejection shapes have tests. The `add_orig` truthiness fix is correct and necessary for v2. Polars `make_origs` got a real rename-crash regression test, and the parity tests were ported with their numeric expectations intact rather than deleted. Deferred-coverage substitutions (PolarsAutocleaning → PandasAutocleaning) are flagged honestly rather than pretending the polars path is exercised.

Local verification before each push: full unit suite 1203 passed, 6 skipped, 0 failed; ruff + `paddy_format --check` clean.

Issue #910, opened 2026-06-10
stat snapshot cache: synchronous writes nearly double cold /load_expr, and the cache layer is unobservable


paddymul:
## Summary

Two related problems with `cache_storage_path` on `/load_expr`, found while profiling a 26,333,291-row entry (buckaroo 0.14.17):

1. Cold `/load_expr` with `cache_storage_path` set takes 7.6s vs 4.1s without it — writing the snapshot parquets nearly doubles the first load, and the cost is paid synchronously inside the request the user is waiting on. (Warm loads drop to 2.9s with no memory spike, so the cache pays off from the second load on.)
2. The cache layer emits no logging at all — no hit, miss, snapshot-written, or write-failed lines. Debugging a live deployment whose stat caches stayed empty (buckaroo-data/tallyman#34 — upstream client timeouts abandoning loads) took hours because nothing on buckaroo's side showed whether snapshots were never requested, attempted and failed, or written somewhere unexpected.

## Output

Measured on the 26.3M-row slice, fresh server per run:

| mode | /load_expr | peak RSS | idle +8s |
|---|---|---|---|
| cache_storage_path set, cold cache | 7.6s | 1,959MB | 88MB |
| cache_storage_path set, warm cache | 2.9s | 358MB | 86MB |
| no cache_storage_path | 4.1s | 2,241MB | 87MB |

## Suggested fixes

- Write snapshots after the response is sent (write-behind / background task), or make synchronous writes opt-in.
- Log one line per pipeline run with cache hits, misses, snapshots written, bytes, and any write errors — write failures in particular must not be silent.

## Context

Same investigation as #906 / #907. Environment: buckaroo 0.14.17, xorq 0.3.26, xorq-datafusion 0.2.7, Python 3.13, macOS arm64.

paddymul:
## Diagnosis update

The synchronous writes aren't what nearly doubles the cold load. The snapshots are stat *results* — a 1-row batch aggregate plus the `<=10`-row histograms (`_categorical_histogram` does `.limit(10)`, the numeric branch is `<=10` buckets) — so each one is 1–4 KB and they don't scale with row count (6 KB total for a 1,000-row table). Moving those writes off the request path would buy almost nothing.

[...]

PR #922, opened 2026-06-11
fix(stats): stream to temp parquet in _create_base_table instead of materializing InMemoryTable

paddymul:
## Problem

`XorqStatPipeline._create_base_table` calls `underlying.create_table(name, table)`, which compiles to a DataFusion `CREATE TABLE AS SELECT`. DataFusion materialises all result rows into a Rust `InMemoryTable`, causing a 12+ GB RSS spike when the stat source is a large filtered parquet scan (e.g. 167M-row parking data with a state filter).

This only fires when `_source_worth_materializing` returns True — plain `Read` nodes are already out-of-core. The spike happens the first time a user applies Search or Sort in the Buckaroo UI.

## Fix

Replace the primary path in `_create_base_table` with:

1. Stream-write the expression to a temp parquet file via `xorq.expr.api.to_parquet` (uses `execute_stream()` internally — DataFusion batches are written incrementally, ~1 GB peak).
2. Register the file as a lazy parquet scan with `underlying.read_parquet(path, table_name=name)`. DataFusion reads columns on demand per stat query.

The temp file path is stored in `self._cache_mat_tmp_parquet` and deleted by `_cleanup_cache_materialization` after the stat run completes. The existing `create_table` fallback chain (in-engine → rewrite CachedNodes → pandas round-trip) is preserved for environments without xorq or when the parquet write fails.

## Verification

- 77 existing `tests/unit/test_xorq_stats_v2.py` tests pass unchanged.
- Plain scan: `_create_base_table` called 0 times (unchanged behaviour).
- NY-filtered 167M-row scan: peak RSS 1.97 GB (was 12+ GB); temp parquet deleted on cleanup.
- `read_parquet` returns `DatabaseTable`, so the `isinstance(materialized.op(), (ops.DatabaseTable, ops.InMemoryTable))` assertion in `test_materializes_past_cached_node` passes.

### DIFF
--- a/buckaroo/pluggable_analysis_framework/xorq_stat_pipeline.py
+++ b/buckaroo/pluggable_analysis_framework/xorq_stat_pipeline.py
@@ -341,14 +341,21 @@ class XorqStatPipeline:
         self._cache_mat_subs = None
         self._cache_mat_cleanup = None
         self._cache_mat_attempted = False
+        self._cache_mat_tmp_parquet = None
 
     def _cleanup_cache_materialization(self):
         cleanup = self._cache_mat_cleanup
+        tmp_parquet = self._cache_mat_tmp_parquet
         self._reset_cache_materialization()
         if cleanup is not None:
             backend, name = cleanup
             try:
                 backend.drop_table(name)
+            except Exception:
+                pass
+        if tmp_parquet is not None:
+            try:
+                tmp_parquet.unlink(missing_ok=True)
             except Exception:
                 pass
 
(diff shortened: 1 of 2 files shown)


paddymul:
## Review + fixes (Claude Code)

Reviewed the streaming-parquet change, verified it behaviorally (schema round-trip, fallback, empty result, CachedNode shape, temp-file lifecycle), then applied the review findings with a failing-CI-first TDD trail. The branch was rebased onto current main (`ba98ccd9`, post dfstats-v1 removal) as part of this; the original fix commit is preserved as `03f17670`.

### Findings and disposition

1. **Silent fallback to the memory-spiking path** — the new `except Exception` swallowed streaming failures with no log, so a failed `to_parquet` (or a future xorq API rename, which the lazy import would hide) silently restores the 12+ GB CTAS behavior. Fixed in `b5fab093`: `log.warning` on fallback, matching the never-silent precedent from #910.

2. **No test pinned the streaming path** — all 77 existing tests also pass without the fix. Added three regression tests in `69fa152f`, committed ahead of the fix so CI shows them failing: [failing run](https://github.com/buckaroo-data/buckaroo/actions/runs/27381370241) (Python / Test 3.11–3.13 + Max Versions), [passing run](https://github.com/buckaroo-data/buckaroo/actions/runs/27381658758) after the fix commits.
   - `test_streaming_path_bypasses_create_table` — spies `create_table`; a worthy source must land via streamed parquet + `read_parquet`, never CTAS.
   - `test_to_parquet_failure_falls_back_with_warning` — stats still correct, warning logged, no partial temp file left.
   - `test_cleanup_releases_table_and_temp_file` — one cleanup call releases both resources.

3. **Split cleanup lifecycle** — the registered table was cleaned via a caller-owned `(backend, name)` tuple while the temp parquet was instance state cleaned elsewhere; a direct `_maybe_materialize` caller (e.g. `test_materializes_past_cached_node`) dropped the table but leaked the file. Fixed in `b5fab093`: `_maybe_materialize` returns just the table and records cleanup in `_cache_mat_cleanup`, so `_cleanup_cache_materialization()` releases table + file together. `process_table`'s finally block simplifies accordingly; direct-caller tests updated.

4. **`_resolve_cached_nodes` fallback now dead in practice** (streaming handles CachedNode exprs, so the rewrite fallback only runs when the parquet write fails, and `test_materializes_past_cached_node`'s docstring describes the old path) — deferred, not addressed here; coverage of the in-engine fallback chain now exists via the forced-failure test, but the CachedNode-specific rewrite branch remains untested when streaming works.

5. **Convention nits** — `tempfile` import moved to module level, redundant `exists()` + `missing_ok` + bare-except collapsed to one guarded `unlink`, docstring corrected (xorq itself is guaranteed by `__init__`; the fallback trigger is a missing `to_parquet` or a failed write). Fixed in `b5fab093`.

6. **tmpfs caveat** — the temp file honours `TMPDIR`; on hosts where /tmp is RAM-backed tmpfs the out-of-core win erodes, so the docstring now says to point `TMPDIR` at real disk for large sources. Documented in `b5fab093`.

Verification: full unit suite green locally (1206 passed, 6 skipped) and all CI checks green on `b5fab093`, including Windows and 3.14.

PR #928, opened 2026-06-16
refactor(dataflow): generic frame typing + basedpyright CI job, clear standard-mode backlog (90→0)

paddymul:
## What

Makes the dataflow pipeline generic over its dataframe type, adds a non-blocking CI job that type-checks the affected files with basedpyright, and clears the standard-mode errors that job reports on those files: **90 on `main` → 0 on this branch**.

### Generic frame typing

New `buckaroo/dataflow/df_types.py` with two type variables:

- `FrameT` (unbounded) for the abstract base. It also covers the lazy `ColumnExecutorDataflow` (`pl.LazyFrame`), which can't meet the eager contract.
- `DataFrameT`, bound to a structural `DataFrameLike` Protocol (`columns` / `__len__` / positional `__getitem__`), for the eager pandas/polars/xorq bodies. No nominal base class spans the three backends.

`ABCDataflow` is `Generic[FrameT]`; `DataFlow` / `CustomizableDataflow` are `Generic[DataFrameT]`. Backends bind concretely: `PandasDataflow`, `PolarsDataflow`, the server dataflows, `ColumnExecutorDataflow(ABCDataflow[pl.LazyFrame])`, and `XorqDataflow(CustomizableDataflow["XorqExpr | pd.DataFrame"])`. The xorq binding is a union because a post-processor may return pandas and a failing one yields a pandas error frame.

traitlets-backed attributes are typed `Any` (a traitlets descriptor returns `Any` on access), and TypedDict annotations on traitlets `Dict` traits are replaced by comments.

### Non-blocking basedpyright CI job

`Python / Typecheck (non-blocking)` runs basedpyright 1.39.8 in standard mode over the files listed in `pyrightconfig.typecheck.json`, on GitHub-hosted runners, with `continue-on-error`. `reportPrivateImportUsage` is off in that config because traitlets re-exports its trait types without `__all__`, which flagged every `from traitlets import ...` line.

### Typing cleanup

- Protocols where backends share a surface but no base class: `Autocleaning` for `ac_obj`, `DfStats` for `DFStatsClass` (`XorqDataflow` casts to `XorqDfStatsV2` at the one site that calls `cache_run_stats()`).
- Column-only helpers (`old_col_new_col`, `merge_sd_overrides`, `merge_column_config`) and `StylingAnalysis` frame arguments take `DataFrameLike`, since polars and xorq frames already reach them.
- Analysis lists are `List[Type[ColAnalysis]]` on the lazy path too (`ColumnExecutorDataflow`, `PAFColumnExecutor`, `PAObjs`, the lazy widget), because the list carries styling classes as well as stat classes.
- Return types fixed on `_df_to_obj`, `_get_summary_sd`, `project_sd` and the infinite-request handlers. `_compute_processed_result` accepts `Optional` since `cleaned_df` is `None` before the first clean.
- `processed_df` modelled as an abstract property; `_summary_sd` None guard; exception reraise guarded against a `None` exc_info.

### Behaviour changes

Positional-only parameters, where a keyword call now raises `TypeError`:

- `BuckarooWidgetBase.add_processing(df_processing_func, /)` and `XorqBuckarooWidget.add_processing(expr_processing_func, /)`. Their parameter names differed. Every caller in buckaroo, the docs and tallyman passes the function positionally (see #963).
- `Sampling.pre_stats_sample(df, /)` and `serialize_sample(df, /)`, so `XorqSampling` can keep its `expr` / `df_or_expr` names. Nothing calls these by keyword.

Everything else is annotation-only.

## basedpyright results (standard mode, scoped files)

| | errors |
|---|---|
| `main`, same file list, default settings | 90 |
| `main`, with `reportPrivateImportUsage` off | 72 |
| this branch | **0** |

The last error, `server/data_loading.py:313` (a pandas slice inferred as `Series | DataFrame`), is fixed by slicing with `.iloc`, which is the same positional slice.

## Test

`pytest tests/unit/`: 1239 passed, 6 skipped locally. The 3 failures (`artifact_test.py::TestToHtml` ×2, `test_mcp_uvx_install.py::test_view_data_call`) need the built JS static files, which this checkout doesn't have; CI builds them.

## Follow-ups

- #963: typed interface for the xorq server path (plugin callable types, `/load_expr` TypedDicts, public `buckaroo.compare` helpers, converting pandas post-processor output to a memtable, a narrower bound for xorq).
- #964: check styling classes' `requires_summary` against the stats the pipeline provides.
- #965: use `StylingAnalysis` in the lazy polars widget.


paddymul:
## xorq typing issues on this branch and where tallyman touches them

Rebased onto main (`0ca27342`) and pushed through `5a998b29`. basedpyright (standard mode, scoped files) is at 4 errors, down from 56 after the rebase. Three of the xorq-related items below are still open or rely on a cast, so this lists each one with the tallyman code path that reaches it.

### 1. `add_processing` parameter name mismatch (open)

`XorqBuckarooWidget.add_processing(self, expr_processing_func)` (`buckaroo/xorq_buckaroo.py:281`) overrides `BuckarooWidgetBase.add_processing(self, df_processing_func)` with a different parameter name. basedpyright reports it twice (on the method and on `XorqBuckarooInfiniteWidget`). At runtime, `widget.add_processing(df_processing_func=f)` on a xorq widget raises `TypeError`; positional calls work. The comment above the method lists the fix options. Nothing is changed yet, pending a check of how callers pass the argument.

**Tallyman exposure:** low. The server path does not call `add_processing`. Project post-processing is loaded by `load_project_post_processing_klasses` (`buckaroo/server/xorq_loading.py:234`), which builds a `ColAnalysis` subclass with `post_process_df` directly (`xorq_loading.py:325`). The marimo export (`tallyman_core/marimo_export.py:115`) constructs `XorqBuckarooInfiniteWidget(_expr, column_config_overrides=...)` and never calls `add_processing`. The only way to hit it is user code in an exported notebook calling `add_processing(df_processing_func=...)`, which already fails on 0.15.3.

Things to grep for on the tallyman branch: `add_processing(` with any keyword argument. If every call is positional, making the base parameter positional-only is the cleanest fix. `tallyman_core/post_processing.py:174` refers to `XorqDataflow.add_processing`; the method lives on `XorqBuckarooWidget`.

### 2. `DFStatsClass` cast in `XorqDataflow._get_summary_sd` (typing only)

`DFStatsClass` is now typed with a `DfStats` Protocol that covers `DfStatsV2`, `PlDfStatsV2` and `XorqDfStatsV2`. `XorqDataflow._get_summary_sd` passes `cache_storage=` and calls `cache_run_stats()`, which only `XorqDfStatsV2` has, so it casts to `type[XorqDfStatsV2]` (`xorq_buckaroo.py:176`). The cast does nothing at runtime. A subclass that set `DFStatsClass` to a non-xorq class would get a `TypeError` on `cache_storage`, as it did before this PR.

**Tallyman exposure:** covered. `XorqServerDataflow` sets `DFStatsClass = XorqDfStatsV2` explicitly (`xorq_loading.py:68`), and the widgets inject it through `InnerDataFlow`. The `cache_storage_path` tallyman passes in `buckaroo_lifecycle.py:601` and `app.py:1177` flows through unchanged.

### 3. Post-processing results and the pandas error frame (annotation only)

`_compute_processed_result` now takes and returns `Optional[DataFrameT]` (`dataflow.py:687`), because `cleaned_df` is `None` until the first clean. When a post-processing function raises, the dataflow still returns a pandas error frame (cast into the frame type), and `XorqDataflow._get_summary_sd` still routes pandas frames to its minimal-SD branch (`xorq_buckaroo.py:163`). `old_col_new_col`, `merge_sd_overrides` and `merge_column_config` are now typed against the `DataFrameLike` protocol, so xorq `Table` and the pandas error frame both type-check. `DataFrameLike.__getitem__` takes its key positionally so `Table` (whose parameter is named `what`) satisfies it.

**Tallyman exposure:** this is the path a project `post_processing/*.py` function takes, including the case where `process()` returns a pandas DataFrame (which `validate_post_processing_source` allows). Worth exercising on the tallyman branch: a post-processing function that returns an expression, one that returns pandas, and one that raises. `_handle_widget_change` now passes the `processed_df` from `widget_args_tuple` to `merge_column_config` instead of re-reading `self.processed_df`. It's the same object, but `tallyman_companion/diff.py` depends on `merge_column_config` applying `column_config_overrides`, so a diff view is a good check too.

### 4. `Sampling` methods are positional-only (the one runtime-visible change)

`Sampling.pre_stats_sample(kls, df, /)` and `serialize_sample(kls, df, /)` (`dataflow_extras.py:63`, `:76`) now take the frame positionally, so `XorqSampling`'s `expr` / `df_or_expr` names are a valid override. A keyword call such as `serialize_sample(df=...)` now raises. Nothing in buckaroo or tallyman calls these by keyword; `XorqInfiniteSampling` inherits them unchanged.

### Pointing tallyman at this branch

- Prefer the TestPyPI build this push will publish (the bot comment above updates with the new `0.15.x.devN` version) over a git dependency. A git source install only builds the JS if `pnpm` is on the path; otherwise hatch writes empty static stubs.
- No wire-protocol changes on main since 0.15.3 (`0.15.3..main` has three commits: #947, #952, #961), so tallyman's `buckaroo-js-core` 0.15.3 pin still matches the renderer side.
- `tests/test_buckaroo_version_lockstep.py::test_buckaroo_python_and_js_pins_match` will fail with a dev pin because the Python and JS versions differ. That failure is expected for this experiment and doesn't indicate drift.

PR #939, opened 2026-06-23
feat(duckdb-node): histogram summary stat (numeric .01–.99 + categorical)

paddymul:
Adds the `histogram` summary stat to `buckaroo-duckdb-node`, the fast-follow #936 reserved (`stats.ts`: "Histograms/quantile displayers are a fast-follow"). A DuckDB column now renders the same histogram bars as the pandas/polars backends, through the existing `displayer: 'histogram'` pinned row.

**Stacked on #936.** Base is `feat/930-duckdb-node-backend` because the package only exists there. Once #936 merges, this retargets to `main` (automatic when the base branch is deleted).

## What's here

- **`histogram.ts`** — a faithful TS port of the Python producer (`customizations/histogram.py`):
  - **Numeric**: clips the meat to the **1st/99th percentile** (`np.quantile` ≡ DuckDB `quantile_cont`), 10 bins, low/high tail markers, NA bucket.
  - **Categorical**: top-7 categories (`cat_pop`), then `longtail` / `unique` / `NA`.
  - 0–100 scale, numpy round-half-to-even, SI-prefix bucket labels.
- **`histogramSql.ts`** — per-column quantile + equal-width-bin SQL (DuckDB has no `width_bucket`, so `floor((v-min)/width)` clamped to the last bin reproduces `np.histogram`'s edges) and the top-N value-counts query, plus the numeric-vs-categorical dispatcher matching `histogram.py:histogram`. A failure on any single column yields an empty histogram, never a broken `initial_state`.
- **`DuckSource.queryRows`** seam (implemented in the node-api adapter) for the small scalar aggregates; the no-coercion row path is untouched.
- **backend** injects the `histogram` row right after `dtype`; **columnConfig** pins it with the histogram displayer (`DefaultMainStyling.pinned_rows` parity).

## Validation

The real-DuckDB spike reproduces numpy's exact normalized meat populations for the Python `INT_ARR` fixture (1st/99th-percentile clipping verified end to end) plus a categorical column. Note CI runs `SKIP_DUCKDB=1`, so the spike runs locally — full suite (77 tests incl. spike) green locally; the non-DuckDB tests run on CI.

## TDD

First commit pushes the contract tests red (they reference the not-yet-added modules); the second commit makes them green.

### DIFF
--- a/packages/buckaroo-duckdb-node/src/DuckSource.ts
+++ b/packages/buckaroo-duckdb-node/src/DuckSource.ts
@@ -26,6 +26,13 @@ export interface DuckSource {
    * natively and the caller never touches a `DuckDBValue`.
    */
   copyToParquet(query: string): Promise<Uint8Array>;
+
+  /**
+   * Run a read query → JSON row objects. Used for the small histogram
+   * aggregates (quantiles, bin counts, value counts), whose results are scalar
+   * counts/labels — not user row data — so the no-coercion rule does not apply.
+   */
+  queryRows(sql: string): Promise<Array<Record<string, unknown>>>;
 }
 
 /** One row of `DESCRIBE (<stmt>)`. DuckDB returns more columns; we use these. */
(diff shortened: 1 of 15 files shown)


paddymul:
@codex review

Issue #943, opened 2026-06-25
telemetry: emit grid-load pipeline timing to the companion (bespoke, session-correlated spans)


paddymul:
tallyman is building a per-project activity/perf log and wants a timeline of a single grid load, but most of that pipeline is invisible outside buckaroo. The companion (tallyman) only sees the first browser request and its own `/load_expr` POST; everything after — summary-stats compute, whether the stat cache was hit, and the WebSocket row-range requests between the browser and the buckaroo server — happens inside buckaroo's server and the `buckaroo-js-core` widget. The row traffic flows browser ↔ server directly over the WS and never reaches the companion, so tallyman is structurally blind to it.

Proposal: instrument the buckaroo **server** and `packages/buckaroo-js-core` (**widget**) to emit OpenTelemetry spans for one grid load, correlated by the buckaroo **session id**, so a consumer (tallyman, or a Jaeger/Tempo backend) can render the waterfall.

## Trace model

One trace per grid load, correlation keyed on the session id minted at `/load_expr`. Spans:

- `load_expr` (server) — POST received → build/replay ready. Continue trace context injected by the caller (tallyman puts it on the POST headers).
- `summary_stats` (server) — start/end, attributes `cache_hit: bool`, `cache_path`, `n_columns`, `n_stats`. (tallyman passes the stat-cache path but never learns hit/miss — only the server knows.)
- `row_range` (server) — one span per row-range request received over the WS: `received_at`, `replied_at`, `start`, `end`, `n_rows`, served-from-row-cache.
- `row_request` / `first_paint` (widget, buckaroo-js-core) — the browser side: when the widget issues its first row request and receives the first rows, plus subsequent requests. Continue the trace over the WS handshake.

## Emission / propagation

- Server: `opentelemetry-sdk` + an OTLP exporter (env-configurable endpoint); accept/continue trace context on the `/load_expr` POST and on the WS connect.
- Widget: `@opentelemetry/api` + a web exporter; propagate context on the WS messages.
- Ingestion (tallyman side, decided at integration): (a) a custom OTLP→`events.jsonl` exporter in the companion so spans land in tallyman's Log UI, and/or (b) a standard collector (Tempo/Jaeger) for the waterfall.

## Why this lives here, not in tallyman

The companion is structurally blind to the WS row traffic and to buckaroo's internal stat compute. Only the server and the widget observe these timings. tallyman captures the request→`load_expr` slice on its own side (Phase 1, tracked in the tallyman repo).

[...]

PR #944, opened 2026-06-25
feat(stats): expose summary-stat cache hit/miss + timing as a structured signal

paddymul:
Part of #943 — the server-side summary-stats cache signal.

## What

The xorq summary-stat snapshot cache already counts hits/misses per `process_table` run, but only emits them as a free-text `log.info` line — nothing structured or timed a telemetry consumer can read.

This makes the outcome a first-class, timed, public signal:

- `XorqStatPipeline` now times each `process_table` run (`_cache_stats["secs"]`).
- New public `XorqStatPipeline.cache_run_stats()` → `{hits, misses, snapshots, bytes, write_errors, secs, cached, status}`, where `status` is `hit` / `miss` / `mixed` / `none` (cache configured) or `uncached`.
- `XorqDfStatsV2.cache_run_stats()` passes it through, so a consumer reading from the stats wrapper (the dataflow / widget) gets it without reaching into `.ap`.
- The existing per-run cache summary log line now includes the elapsed time.

## Why

#943 wants a timeline of the grid-load pipeline. "Whether the stat cache was used for summary stats" and "summary-stats start/end" are two of the requested signals, and they're only observable inside buckaroo — the companion (tallyman) passes the cache path but never learns hit/miss. This exposes that signal as structured, timed data: the foundation an OTel emitter (or any consumer) attaches to.

## Scope — not in this PR

The telemetry **transport** — OTel SDK + trace-context propagation across companion → server → widget, plus the WebSocket row-range spans — remains the larger #943 work. This PR is the server-side cache/stats signal made first-class; it changes no request/response contract and adds no dependency.

## Tests

`tests/unit/test_xorq_stats_v2.py::TestSnapshotCacheRun`:
- `test_cache_run_stats_reports_status_and_timing` — cold run → `status="miss"` + `secs`; warm run → `status="hit"`.
- `test_cache_run_stats_uncached` — no cache → `status="uncached"`.

70 stats tests pass; ruff + paddy-format clean (pre-commit + pre-push).

### DIFF
--- a/buckaroo/pluggable_analysis_framework/perf_log.py
+++ b/buckaroo/pluggable_analysis_framework/perf_log.py
@@ -29,11 +29,15 @@ stays quiet unless enabled, regardless of how noisy other loggers are.
 """
 from __future__ import annotations
 
+import json
 import logging
 import os
 import time
+import urllib.request
+from concurrent.futures import Future, ThreadPoolExecutor
 from contextlib import contextmanager
-from typing import List, Tuple
+from contextvars import ContextVar
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 log = logging.getLogger("buckaroo.perf")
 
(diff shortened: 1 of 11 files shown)


paddymul:
## Review pass (`/review`) — findings + dispositions

A medium-effort review of this PR surfaced 6 findings; here's how each was handled. Bugfixes followed the test-first-red-on-CI flow.

**TDD proof:**
- Failing tests: `88dc3fb7` → [red on CI](https://github.com/buckaroo-data/buckaroo/actions/runs/28206323327) (the xorq jobs — 3.11/3.12/3.13 — failed; 3.14 passed because xorq isn't installed there so the tests skip)
- Fixes: `c36b1632` + `827b402a` → [green on CI](https://github.com/buckaroo-data/buckaroo/actions/runs/28206715756) (all Python test jobs pass)

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | **Warm-session early-exit drops telemetry.** A `/load_expr` re-POST matching a warm session (same `build_dir`, no `force_reload`/config — `telemetry_url` isn't a config-bearing field) returned cached metadata *before* the sink was built and *before* `_perf_first_payload_seen` was reset, so a page-refresh's WS first-pull emitted nothing and a fresh `telemetry_url` was ignored. | **Fixed** `c36b1632` — build the sink before the early-exit; in the warm branch rebind `existing.tele_sink` and reset the first-pull flag, so the warm path re-arms exactly like the full-load path. Test: `test_first_pull_telemetry_rearms_after_warm_reload`. |
| 2 | **`_span` decoupling gap.** `XorqStatPipeline._span` gated solely on `perf_log.enabled()`, so a telemetry-only run (`BUCKAROO_PERF` off, sink bound) emitted `firstpull.*` but never the `stat.xorq.*` timeline. | **Fixed** `c36b1632` — defer the gate to `perf_span` (live on enabled-OR-sink), keeping only the unit-test suppression in `_span`. Test: `test_internal_spans_emit_to_telemetry_sink_with_perf_off`. |
| 3 | **First-pull flag set only on success → double emit on dispatch-retry.** | **No change.** The failed first dispatch is still logged (`log.error`) *and* emitted as an `errored=true` span record — nothing is lost. Leaving the flag unset is intentional (keep emitting until a clean first-pull lands); the `errored=true` tag is the dedupe key for a consumer. |
| 4 | **`make_http_sink` docstring overstates backpressure** — `max_clients` caps concurrency, not queue depth; a dead companion under sustained emission queues unboundedly. | **Filed** as #946 (latent; current `firstpull.*`-only volume keeps it dormant). |
| 5 | **Duplicated first/second WS payload blocks.** | **Fixed (refactor)** `827b402a` — extracted `_dispatch_and_send(pa, span_label)`; behavior-preserving, covered by the existing WS payload telemetry tests. |
| 6 | **`cache_run_stats()` computed unconditionally** on the no-op span path. | **Declined** — negligible (a dict copy + branch); not worth a guard. |

paddymul:
@codex review

